PR #8, opened 2022-09-01
Upload packages to PyPI

### LINKED ISSUE #1: Uploads packages to pypi
Resources:
* Tutorial packaging with a lot of good resources: https://python-packaging-tutorial.readthedocs.io/en/latest/uploading_pypi.html
* Signing packages: https://twine.readthedocs.io/en/stable/#why-should-i-use-this
* Example repository from PyPA (Python Package Authority): https://github.com/pypa/sampleproject
* Another simple but real-life example project: https://github.com/pypa/build
* Summary on different packaging/build tools: https://towardsdatascience.com/packaging-in-python-tools-and-formats-743ead5f39ee#fb39

leandro-lucarella-frequenz:
This PR does a bit of cleanup and modernization of the project and upload
packages to PyPI when new tags are pushed.

- Exclude GitHub and test files from MANIFEST.in
- Move project metadata/config to pyproject.toml
- Add extra metadata to pyproject.toml
- Move build-py-proto to setup.py
- Don't run the CI GitHub workflow on release
- Publish distribution files to PyPI


leandro-lucarella-frequenz:
I already tested this in my fork, here are the produced packages: https://pypi.org/project/frequenz-api-microgrid/#history (these are tags I created in my forks for testing purposes).

.github/workflows/ci.yaml
```diff
@@ -52,10 +51,53 @@ jobs:
...
+    steps:
+      - name: Fetch sources
+        uses: actions/checkout@v3
+        with:
+          submodules: true
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
```
tiyash-basu-frequenz, line 70: Asking as a python noob:
Is building it with v3.10 enough? Do we need to build it with other python versions too?

leandro-lucarella-frequenz:
I added one commit to update to the new repository name.

.github/workflows/ci.yaml
leandro-lucarella-frequenz: I'm not a 100% sure but the wheel doesn't carry information about the python version (only if it's python 2 or 3, for example: `frequenz_api_microgrid-0.11.0a1-py3-none-any.whl`). Since we don't have different code for different python versions, I don't see how making different packages could make any sense, we just need to make sure our **single code base** stays compatible with all Python versions we want to support.

I will try to verify this though.

tiyash-basu-frequenz: Okay. I think it's okay if we get this in, and fix things in a future PR, if required at all.

tiyash-basu-frequenz (approved):
LGTM!

.github/workflows/ci.yaml
leandro-lucarella-frequenz: Confirmed: we **can** ship different wheels for different python versions (if we use, for example `-py37`) but we **don't have to** (because as I said we support all the different Python versions through the same code base). The meaning of `-py3` is *any Python 3 version* (then the metadata inside the wheel can narrow this down).

Some useful links for context:
* Wheel specs: https://packaging.python.org/en/latest/specifications/binary-distribution-format/
* Wheel PEP: https://peps.python.org/pep-0427/
* Compatibility Tags for Built Distributions: https://peps.python.org/pep-0425/#python-tag

PR #9, opened 2022-09-05
ci: Create GitHub Release automatically

leandro-lucarella-frequenz:
Add a GitHub job to create the GitHub Release automatically when a version is tagged. Also use the GitHub feature to create the `CHANGELOG` automatically. For this the `CHANGELOG.md` is removed in favor of a `RELEASE_NOTES.md` files that should only contain an explanation on the release main changes.

Also add the release notes for the current release.

### DIFF
--- a/.github/RELEASE_NOTES.template.md
+++ b/.github/RELEASE_NOTES.template.md
@@ -1,0 +1,17 @@
+# Frequenz Migrogrid API Release Notes
+
+## Summary
+
+<!-- Here goes a general summary of what this release is about -->
+
+## Upgrading
+
+<!-- Here goes notes on how to upgrade from previous versions, including if there are any depractions and what they should be replaced with --> 
+
+## New Features
+
+<!-- Here goes the main new features and examples or instructions on how to use them -->
+
+## Bug Fixes
+
+<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -84,9 +84,9 @@ jobs:
           path: dist/
           if-no-files-found: error
 
-  publish-to-pypi:
+  create-github-release:
     needs: ["protolint", "tests", "build-dist"]
-    # Publish only on tags creation
+    # Create a release only on tags creation
     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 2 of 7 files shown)


leandro-lucarella-frequenz:
This PR is also a RFC because it changes the way we build the CHANGELOG. I noticed there is a lot of manual work while GitHub is providing the same automatically for free, so what I'm proposing is to stop wringing a `CHANGELOG.md` and write a `RELEASE_NOTES.md` with a more high-level description of what was mainly changed in the release. Here we should include steps needed to migrate from previous versions, examples on new features, etc.

This is how the generated release notes look like:

https://github.com/leandro-lucarella-frequenz/frequenz-api-microgrid-proto/releases/tag/untagged-45de1f171a0afe9f6be2

Everything above `What's Changed` is the contents of `RELEASE_NOTES.md` and everything below it was automatically generated by GitHub.

The main advantage is to reduce manual work (we are basically doing the same GH does automatically by listing PRs), the downside is there is no static CHANGELOG (we use the releases page from GH as CHANGELOG).

leandro-lucarella-frequenz:
ping @tiyash-basu-frequenz @matthias-wende-frequenz @ela-kotulska-frequenz because I'm suggesting a change in the way we do releases and write changelogs / release notes.

sahas-subramanian-frequenz:
So we create tags locally and push?  Or is there a way to create tags on gh without a release?

.github/RELEASE_NOTES.template.md
```diff
@@ -0,0 +1,13 @@
...
+
+<!-- Here goes a general summary of what this release is about -->
+
+## Upgrading
+
+<!-- Here goes notes on how to upgrade from previous versions, including if there are any depractions and what they should be replaced with --> 
+
+## New Features
+
+<!-- Here goes a the main new features and examples or instructions on how to use them -->
```
tiyash-basu-frequenz: Is it possible to add a section for bug fixes here?

leandro-lucarella-frequenz:
> So we create tags locally and push? Or is there a way to create tags on gh without a release?

Yeah, I just added the steps to release to the guide. Just tag and push.

.github/RELEASE_NOTES.template.md
leandro-lucarella-frequenz: I thought these might be details for the extended changelog, but we can add it and if there is nothing that needs special mention, we can remove the section before releasing :+1:

leandro-lucarella-frequenz: The main point here is not just duplicating the whole list of PRs in this section, we should only mention stuff that need some special explanation.

leandro-lucarella-frequenz: Added

leandro-lucarella-frequenz:
> So we create tags locally and push? Or is there a way to create tags on gh without a release?

I didn't thought about creating release+tag via the GH UI but I considered create a release in GitHub and trigger the uploading of the packages to PyPI from that event, but always thinking about pushing a tag first. I found it more complicated, because we should either repeat all the tests for the release event or find a way to wait until the tag is tested to make the release. I thought this workflow was simpler: tag triggers -> create release and upload packages.

If both creating the GH release and the tag are done in the UI I think is simpler, but maybe a bit less resilient if someone wants to create the tag first from the CLI or something.

tiyash-basu-frequenz (approved):
LGTM!

PR #17, opened 2022-09-12
Remove unnecessary setuptools.find configuration

leandro-lucarella-frequenz:
When configuring through pyproject.toml, setuptools will automatically discover packages using native namespaces in the package-dir.


tiyash-basu-frequenz (approved):
LGTM

PR #34, opened 2023-04-12
Specify non-guarantee of 1W resolution in methods

### LINKED ISSUE #23: Document that control function calls don't guarantee watt resolution
### What's needed?

`Charge` and `Discarche` functions, for example, take a power in watt as a `int64` (and will probably even be updated to float in the future), but not all devices support this fine-grained resolution.

### Proposed solution

Document this behavior in the documentation.

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
Some components may have a charge/discharge power resolution of more than 1 W. E.g., an inverter may have a resolution of 88 W. In such cases, the charge/discharge power will be floored to the nearest multiple of the resolution.

### DIFF
--- a/proto/frequenz/api/microgrid/microgrid.proto
+++ b/proto/frequenz/api/microgrid/microgrid.proto
@@ -97,6 +97,11 @@ service Microgrid {
   // Sets the charge power of a component with a given ID, provided the
   // component supports it.
   //
+  // Note that the target component may have a resolution of more than 1 W.
+  // E.g., an inverter may have a resolution of 88 W.
+  // In such cases, the charge power will be floored to the nearest multiple of
+  // the resolution.
+  //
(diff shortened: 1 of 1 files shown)


sahas-subramanian-frequenz:
Rounding issues was why we were using int instead of float for power values right?  If we're rounding anyway,  what do you think about switching to float, to have the same type for data and control?

tiyash-basu-frequenz:
>what do you think about switching to float, to have the same type for data and control?

Why not. I will add the change here: https://github.com/frequenz-floss/frequenz-api-microgrid/pull/36
These existing methods will become deprecated anyway.

tiyash-basu-frequenz:
Signed the commit to fix the DCO issue.

sahas-subramanian-frequenz (approved):

PR #38, opened 2023-04-12
Support returning max allowed current at grid connection point

### LINKED ISSUE #19: Report grid connection point's current limit
### What's needed?

Every microgrid has a limit on how much current is allowed to flow through the grid connection point. This is generally a static value, dictated by the fuses/circuit breakers placed in the microgrid.

The microgrid API should report this value, so that clients can incorporate it into their decision making mechanisms.

### Proposed solution

_No response_

### Use cases

This value can be used by clients to preemptively check if a command does not allow the current at a grid connection point to exceed its fuse's limits.

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
This PR
* adds component-category-specific metadata, and
* introduces metadata for the grid component category, and
* adds the max allowed current to the grid metadata

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -28,6 +28,12 @@
   (-ve for production, and +ve for consumption). This new method allows setting
   active power values in the same convention, making the API more consistent.
 
+* [Introduced component category-specific metadata](https://github.com/frequenz-floss/frequenz-api-microgrid/pull/36).
+  This metadata is returned in response to `ListComponents` calls,
+  in a new message variable `Component.metadata`.
+  This is a more general way of representing category-specific metadata,
+  like category-type, and removes `Component.type`.
+
 ## Bug Fixes
 
 <!-- Here goes notable bug fixes that are worth a special mention or explanation -->
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -34,6 +34,15 @@
   This is a more general way of representing category-specific metadata,
   like category-type, and removes `Component.type`.
 
+* [Introduced grid max-current](https://github.com/frequenz-floss/frequenz-api-microgrid/pull/38).
+  This change introduces a `grid.Metadata` message, which contains the item
+  `rated_fuse_current`. This is the rating of the fuse at the grid connection
+  point.
+  This rating specifies the maximum amount of current, measured in amperes,
+  that can flow in or out of each of the 3 phases individually.
+  The current _i_ A at the grid connection point must comply with the
+  following constraint: : `-rated_fuse_current <= i <= rated_fuse_current`
+
 ## Bug Fixes
 
 <!-- Here goes notable bug fixes that are worth a special mention or explanation -->
--- a/proto/frequenz/api/microgrid/grid.proto
+++ b/proto/frequenz/api/microgrid/grid.proto
@@ -1,0 +1,26 @@
+// Contains grid connection point details.
+//
+// Copyright:
+// Copyright 2023 Frequenz Energy-as-a-Service GmbH
+//
+// License:
+// MIT
+
+syntax = "proto3";
+
+package frequenz.api.microgrid.grid;
+
+// The grid connection point metadata.
+message Metadata {
+  // This refers to the maximum amount of electrical current, in amperes, that a
+  // fuse at the grid connection point is designed to safely carry under normal
+  // operating conditions.
+  //
+  // This limit applies to currents both flowing in or out of each of the 3
+  // phases individually.
+  //
+  // In other words, a current _i_ A at one of the phases of the grid connection
+  // point must comply with the following constraint:
+  // `-rated_fuse_current <= i <= rated_fuse_current`
+  uint32 rated_fuse_current = 1;
+}
--- a/proto/frequenz/api/microgrid/microgrid.proto
+++ b/proto/frequenz/api/microgrid/microgrid.proto
@@ -412,15 +412,13 @@ message Component {
   // The model name of the component.
   string model_name = 5;
 
-  // The type of the component.
-  // E.g., if the `ComponentCategory` is `INVERTER`, then the inverter type
-  // (`UNSPECIFIED`, `BATTERY`, `SOLAR`, or `HYBRID`)
-  oneof type {
-    battery.Type battery = 6;
-    inverter.Type inverter = 7;
-    meter.Type meter = 8;
-    ev_charger.Type ev_charger = 9;
-    sensor.Type sensor = 10;
+  // The metadata specific to the component category.
+  oneof metadata {
+    battery.Metadata battery = 11;
+    inverter.Metadata inverter = 12;
+    meter.Metadata meter = 13;
+    ev_charger.Metadata ev_charger = 14;
+    sensor.Metadata sensor = 15;
   }
 }
 
--- a/proto/frequenz/api/microgrid/microgrid.proto
+++ b/proto/frequenz/api/microgrid/microgrid.proto
@@ -13,6 +13,7 @@ package frequenz.api.microgrid;
 import "frequenz/api/microgrid/battery.proto";
 import "frequenz/api/microgrid/common.proto";
 import "frequenz/api/microgrid/ev_charger.proto";
+import "frequenz/api/microgrid/grid.proto";
 import "frequenz/api/microgrid/inverter.proto";
 import "frequenz/api/microgrid/meter.proto";
 import "frequenz/api/microgrid/sensor.proto";
@@ -414,11 +415,12 @@ message Component {
 
   // The metadata specific to the component category.
   oneof metadata {
-    battery.Metadata battery = 11;
-    inverter.Metadata inverter = 12;
-    meter.Metadata meter = 13;
-    ev_charger.Metadata ev_charger = 14;
-    sensor.Metadata sensor = 15;
+    grid.Metadata grid = 11;
+    battery.Metadata battery = 12;
+    inverter.Metadata inverter = 13;
+    meter.Metadata meter = 14;
+    ev_charger.Metadata ev_charger = 15;
+    sensor.Metadata sensor = 16;
   }
 }
 
--- a/proto/frequenz/api/microgrid/battery.proto
+++ b/proto/frequenz/api/microgrid/battery.proto
@@ -27,6 +27,12 @@ enum Type {
   TYPE_NA_ION = 3;
 }
 
+// The battery metadata.
+message Metadata {
+  // The battery type.
+  Type type = 1;
+}
+
 enum ComponentState {
   // Default value.
   COMPONENT_STATE_UNSPECIFIED = 0;
--- a/proto/frequenz/api/microgrid/ev_charger.proto
+++ b/proto/frequenz/api/microgrid/ev_charger.proto
@@ -28,6 +28,12 @@ enum Type {
   TYPE_HYBRID = 3;
 }
 
+// The EV charger metadata.
+message Metadata {
+  // The EV charger type.
+  Type type = 1;
+}
+
 // The possible states of an EV charging station.
 enum ComponentState {
   // Default state.
--- a/proto/frequenz/api/microgrid/inverter.proto
+++ b/proto/frequenz/api/microgrid/inverter.proto
@@ -20,6 +20,12 @@ enum Type {
   TYPE_HYBRID = 3;
 }
 
+// The inverter metadata.
+message Metadata {
+  // The inverter type.
+  Type type = 1;
+}
+
 // Enumerated inverter states.
 enum ComponentState {
   // Default value.
--- a/proto/frequenz/api/microgrid/meter.proto
+++ b/proto/frequenz/api/microgrid/meter.proto
@@ -29,6 +29,12 @@ enum Type {
   TYPE_PURCHASED = 4;
 }
 
+// The meter metadata.
+message Metadata {
+  // The meter type.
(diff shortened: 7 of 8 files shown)


tiyash-basu-frequenz:
An alternative would have been to just add a `max_current_magnitude` item to the `Component` message, but that would have introduced two different methods of setting bounds on inverters and batteries (the other one being dynamic bounds). Would be nice to hear your thoughts on this idea as well, besides the PR.

proto/frequenz/api/microgrid/grid.proto
```diff
@@ -0,0 +1,19 @@
...
+syntax = "proto3";
+
+package frequenz.api.microgrid.grid;
+
+// The grid connection point metadata.
+message Metadata {
+  // The maximum current magnitude in amperes.
+  // This is the maximum current that can be drawn from or supplied to
+  // the grid.
+  uint32 max_current_magnitude = 1;
```
thomas-nicolai-frequenz: Magnitude? I've never experienced us using this term so far. Why couldn't it just be name `max_current` instead?

thomas-nicolai-frequenz: If we wanna introduce this term here we need to describe it better in the comment what that is.

tiyash-basu-frequenz: The maximum current at a fuse is a signed value. Meaning, if the fuse limit is 5A, then a current _i_ A should be `-5 <= i <= 5`, or `|i| <= 5`. Hence I added the term `magnitude`, to indicate that this limit applies to +ve as well as -ve currents.

Or do we want to split this? Then we could specify that at a site we can consume from the grid, but never supply to it.

tiyash-basu-frequenz: >Or do we want to split this? Then we could specify that at a site we can consume from the grid, but never supply to it.

I made the changes.

thomas-nicolai-frequenz: > Hence I added the term magnitude, to indicate that this limit applies to +ve as well as -ve currents.

Well thats where the inconsistency comes in. We have signed values in many places and yet don't call them all `magnitudes`. I think its enough to have that in the description tbh otherwise we would need to change this all over the place.

thomas-nicolai-frequenz: > I made the changes

Don't! This is the fuse limit. A fuse imho doesn't care in which direction the power is flowing. This I would call this `max_grid_fuse_current` or something tbh. I would suggest to later introduce virtual limits for grid feed-in or grid consumption once we need that. More likely we will add frequency limits later on as the first step but lets postpone that discussion until we really need it.

tiyash-basu-frequenz: Done. I renamed the variable to `max_grid_fuse_current`.

RELEASE_NOTES.md
```diff
@@ -28,6 +28,12 @@
...
+* [Introduced component category-specific metadata](https://github.com/frequenz-floss/frequenz-api-microgrid/pull/36).
+  This metadata is returned in response to `ListComponents` calls,
+  in a new message variable `Component.metadata`.
+  This is a more general way of representing category-specific metadata,
+  like category-type, and deprecates `Component.type`.
+
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+<!-- Here goes notable bug fixes that are worth a special mention or explanation -->`
```
leandro-lucarella-frequenz: I guess this change is not intended?

proto/frequenz/api/microgrid/microgrid.proto
```diff
@@ -412,15 +413,14 @@ message Component {
...
-  // (`UNSPECIFIED`, `BATTERY`, `SOLAR`, or `HYBRID`)
-  oneof type {
-    battery.Type battery = 6;
-    inverter.Type inverter = 7;
-    meter.Type meter = 8;
-    ev_charger.Type ev_charger = 9;
-    sensor.Type sensor = 10;
+  // The metadata specific to the component category.
+  oneof metadata {
+    grid.Metadata grid = 10;
```
leandro-lucarella-frequenz: Shouldn't this start at 11 to avoid repeating used tags?

Maybe also add a comment here saying which tags were used in the past and should not be reused?

leandro-lucarella-frequenz (commented):
Other than the few comments, LGTM.

RELEASE_NOTES.md
tiyash-basu-frequenz: oops, looks like a typo. Will fix.

proto/frequenz/api/microgrid/microgrid.proto
tiyash-basu-frequenz: > Shouldn't this start at 11 to avoid repeating used tags?

I don't think repeating used tags is a concern, because this is going to be rolled out as a breaking change. But if you really want new tags, I can update these.

RELEASE_NOTES.md
tiyash-basu-frequenz: Fixed.

proto/frequenz/api/microgrid/grid.proto
```diff
@@ -0,0 +1,22 @@
...
+
+// The grid connection point metadata.
+message Metadata {
+  // The rating of the fuse at the grid connection point, in amperes.
+  //
+  // Note that this limit applies to the magnitude of the current flowing both
+  // into or out of the grid connection point. E.g., a current _i_ A at the grid
+  // connection point has to follow the constraint:
+  // -max_grid_fuse_current <= i <= max_grid_fuse_current
+  uint32 max_grid_fuse_current = 1;
```
leandro-lucarella-frequenz: Also minimal, but since this is already in the scope of the grid component category maybe we can just leave it at `max_fuse_current`.

tiyash-basu-frequenz: Done

tiyash-basu-frequenz:
Updated the docs to specify that the limit applies individually to all the 3 phases.

proto/frequenz/api/microgrid/microgrid.proto
leandro-lucarella-frequenz: OK, it was strange that you use new tags for all the metadata except for the new `grid` one, so it looked like something not intentional.

In any way, IIRC, it is a good practice to not reuse tags in protobufs, as at the protocol level if we do if for some reason we have mismatched versions between the server and the client, and the server is sending some `grid` metadata the client might interpret it as some old `sensor` type instead, maybe making the client code blow up in a more unexpected way.

I agree with such big change it will probably blow up pretty soon and consistently but if the changes were smaller it could be a harder to find issue if versions are accidentally mixed.

In any case, I'd strongly prefer to either:
* Not reuse tags
* Keep tags continuous

Right now it is weird that we suddenly jump from 5 to 10 for no apparent reason (if you don't see this PR :).

tiyash-basu-frequenz: >so it looked like something not intentional.

I just did it to keep `10` available for `grid` in the 2nd commit :)

> at the protocol level if we do if for some reason we have mismatched versions between the server and the client, and the server is sending some grid metadata the client might interpret it as some old sensor type instead, maybe making the client code blow up in a more unexpected way.

Yes, that will happen, and I agree that we should not reuse tags (in fact it's a standard practice to not reuse tags).
But we are at a very early stage, and doing it now will blow up the tag numbers very quickly. I'd suggest waiting until we release v1.0.0 until we introduce such constraints. 

> Right now it is weird that we suddenly jump from 5 to 10 for no apparent reason (if you don't see this PR :).

True. We have left intentional gaps in places so that we can introduce new items in a non-breaking change.
But this is not required, since the numbers can be arbitrary. We can fix those in a cleanup PR.

tiyash-basu-frequenz: I'll just go ahead an renumber the tags, so that older ones are not used.

... and done.

leandro-lucarella-frequenz: OK, as I said before, for me consistency/information is the most important. If you want to reuse tags because it is very early in the development, then sure, but I would clarify in the commit message for example, and chose contiguous tags.

If there is a gap in the tags, either add a comment in the code saying the gap is there because there were some old attributes using those tags so they shouldn't be reused, or maybe have it as a general rule of the project that gaps can't be reused (but IMHO being explicit locally is always better and less error prone).

leandro-lucarella-frequenz (dismissed):
As said in the previous comment, I would add a comment about the gap in the code for extra clarity, but not a blocker, so approving.

proto/frequenz/api/microgrid/microgrid.proto
tiyash-basu-frequenz: Fair enough 👍

tiyash-basu-frequenz:
Following our discussion [here](https://github.com/frequenz-floss/frequenz-sdk-python/discussions/322#discussioncomment-5603799), I have updated the PR to introduce a `Fuse` message with limits for individual phases, and added it to the `grid.Metadata` message

proto/frequenz/api/microgrid/grid.proto
sahas-subramanian-frequenz: is this checked in?  I still see 3 vars, one for each phase.

tiyash-basu-frequenz: Not yet. I am waiting for further input from Niklas. I will update the PR once I have all the details.

tiyash-basu-frequenz:
I reverted the last change according to the latest discussions and inputs.

Now we just have the following:
```proto
// The grid connection point metadata.
message Metadata {
  // The rating of the fuse at the grid connection point.
  // This rating specifies the maximum amount of current, measured in amperes,
  // that can flow in or out of each of the 3 phases individually.
  // The current _i_ A at the grid connection point must comply with the
  // following constraint: : `-max_fuse_current <= i <= max_fuse_current`
  uint32 max_fuse_current = 1;
}

```

thomas-nicolai-frequenz:
> max_fuse_current

Shall we call it `fuse_current_limit` instead? the term max feels a little ambiguous.

leandro-lucarella-frequenz:
`fuse_limit_current`? So we can change the suffix if there are new limits.

It is quite unfortunate that no matter what it is still ambiguous and could be misunderstood as the limit *rigth now* instead of the limit on the electrons flow.

tiyash-basu-frequenz:
How about `rated_fuse_current`? The inclusion of the term `fuse` itself implies that the current should be limited here.

thomas-nicolai-frequenz:
> It is quite unfortunate that no matter what it is still ambiguous and could be misunderstood as the limit rigth now instead of the limit on the electrons flow.

true.

tiyash-basu-frequenz:
>How about `rated_fuse_current`? The inclusion of the term fuse itself implies that the current should be limited here. 

Done. Ready for another look.

proto/frequenz/api/microgrid/grid.proto
tiyash-basu-frequenz: Renamed it to `rated_fuse_current`.

https://github.com/frequenz-floss/frequenz-api-microgrid/pull/38#issuecomment-1514732619

RELEASE_NOTES.md
```diff
@@ -28,6 +28,21 @@
...
+  like category-type, and deprecates `Component.type`.
+
+* [Introduced grid max-current](https://github.com/frequenz-floss/frequenz-api-microgrid/pull/38).
+  This change introduces a `grid.Metadata` message, which contains the item
+  `max_fuse_current`. This is the rating of the fuse at the grid connection
+  point.
+  This rating specifies the maximum amount of current, measured in amperes,
+  that can flow in or out of each of the 3 phases individually.
+  The current _i_ A at the grid connection point must comply with the
+  following constraint: : `-max_fuse_current <= i <= max_fuse_current`
```
leandro-lucarella-frequenz: This will need to be updated with the new name.

RELEASE_NOTES.md
```diff
@@ -28,6 +28,21 @@
   (-ve for production, and +ve for consumption). This new method allows setting
   active power values in the same convention, making the API more consistent.
 
+* [Introduced component category-specific metadata](https://github.com/frequenz-floss/frequenz-api-microgrid/pull/36).
+  This metadata is returned in response to `ListComponents` calls,
+  in a new message variable `Component.metadata`.
+  This is a more general way of representing category-specific metadata,
+  like category-type, and deprecates `Component.type`.
```
leandro-lucarella-frequenz: This is actually **removing**, not deprecating.

leandro-lucarella-frequenz (commented):
Other than a few comments about outdated docs, it looks like we have a winner!

RELEASE_NOTES.md
tiyash-basu-frequenz: Fixed 🚀

RELEASE_NOTES.md
tiyash-basu-frequenz: Fixed 🚀

leandro-lucarella-frequenz (approved):
Unless there is any pending topic in the SDK discussion that need to be considered here, this is, as far as I am concerned, ready to go! :shipit:

tiyash-basu-frequenz:
@thomas-nicolai-frequenz Do you have any further input on this? If not, we can get this merged.

thomas-nicolai-frequenz:
I think I have created enough trouble :-) Lets merge it!

PR #39, opened 2023-04-13
Introduce minimum bounds

### LINKED ISSUE #28: Introduce minimum bounds for active power
### What's needed?

The API supports setting maximum bounds for active power, but not for setting minimum bounds. The API should have the ability to set minimum bounds.

### Proposed solution

The [`Metric`](https://github.com/frequenz-floss/frequenz-api-microgrid/blob/e1380cb7b567588ff9fbdca32c8c935d17bd5314/proto/frequenz/api/microgrid/common.proto#L23-L39) and the [`MetricAggregation`](https://github.com/frequenz-floss/frequenz-api-microgrid/blob/e1380cb7b567588ff9fbdca32c8c935d17bd5314/proto/frequenz/api/microgrid/common.proto#L48-L72) messages contain maximum bounds as of now. These need to be refactored for having minimum bounds as well.

Also, the [`SetBoundsParam`](https://github.com/frequenz-floss/frequenz-api-microgrid/blob/v0.x.x/proto/frequenz/api/microgrid/microgrid.proto#L301-L316) message has to be refactored to allow adding minimum bounds.

### Use cases

This is useful when we want to prevent agents from charging or discharging with too low power. This can be used, e.g., to limit minimum charging power to a high enough value so that a peak-shaver agebt does not interfere with the DC heater agent.

_No response_

### Additional context

_No response_

### LINKED ISSUE #35: Add a simple RPC for setting bounds
### What's needed?

There is an existing client-side streaming method for setting bounds:
https://github.com/frequenz-floss/frequenz-api-microgrid/blob/506bfa4a35a13e59568bfea8e9378b79a8612a84/proto/frequenz/api/microgrid/microgrid.proto#L94-L95

However, for some clients, e.g., command-line ones, it is not not straightforward to perform client-side streaming. ATM, such clients create the stream, send one `SetBoundsParam` through it, and end the call.

A simpler solution would be to add a simple RPC to set one-off bounds, which accepts just one `SetBoundsParam` object, and returns a timestamp until when the said bounds object would stay in effect.

Note: "A simple RPC where the client sends a request to the server using the stub and waits for a response to come back, just like a normal function call." (from https://grpc.io/docs/languages/go/basics/).

### Proposed solution

The proposed solution would be to add the following RPC:

```proto
  rpc SetBounds(SetBoundsParam) returns (google.protobuf.Timestamp);
```

As mentioned above, the return value would be the timestamp until the provided bounds would stay in effect.

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
In the message `common.Metric`, `system_bounds` has now been replaced by `system_bounds_min` and `system_bounds_max`. A metric value `v` should now follow the following constraint:
  `v >= system_bounds_max.lower` AND `v <= system_bounds_min.lower` AND `v >= system_bounds_min.upper` AND `v <= system_bounds_max.upper`

  Also, two new simple RPCs have been added:
  * `AddMinBounds`: adds a minimum bound pair, and returns the UTC timestamp until when it will stay in effect.
  * `AddMaxBounds`: adds a maximum bound pair, and returns the UTC timestamp until when it will stay in effect.

  and the RPC `SetBounds` has been deprecated.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -67,6 +67,24 @@
   `----` values here are disallowed and wil be rejected, and
   `====` values here are allowed and will be accepted.
 
+  Two new simple RPCs for setting exclusion and inclusion bounds have been
+  added:
+  * `AddExclusionBounds`: adds a pair of exclusion bounds for a given component
+    and metric, and returns the UTC timestamp until when it will stay in effect.
+  * `AddInclusionBounds`: adds a pair of inclusion bounds for a given component
+    and metric, and returns the UTC timestamp until when it will stay in effect.
+
+  Exclusion bounds are a useful tool for enhancing the performance of a system.
+  They can be used to restrict the acceptance of commands that fall below a
+  certain threshold, which can help ensure the smooth functioning of the system.
+  E.g., exclusion bounds can be set to limit the minimum charging power to a
+  sufficiently high level, preventing a peak-shaver client from sending charge
+  powers that are too low when a DC heater client is executing a charge pulse.
+  This can significantly improve the overall performance of the DC heating
+  mechanism.
+
+  The RPC `SetBounds` has been deprecated.
+
 ## Bug Fixes
 
 <!-- Here goes notable bug fixes that are worth a special mention or explanation -->
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -43,6 +43,30 @@
   The current _i_ A at the grid connection point must comply with the
   following constraint: : `-rated_fuse_current <= i <= rated_fuse_current`
 
+* [Introduced exclusion bounds](https://github.com/frequenz-floss/frequenz-api-microgrid/pull/39).
+  In the messages `common.Metric` and `common.MetricAggregation`,
+  `system_bounds` has now been replaced by `system_exclusion_bounds` and
+  `system_inclusion_bounds`. A metric's `value` now has to comply with the
+  following constraints:
+
+  * `value <= system_exclusion_bounds.lower` OR
+    `system_exclusion_bounds.upper <= value`
+
+  * `system_inclusion_bounds.lower <= value <= system_inclusion_bounds.upper`
+
+  `system_inclusion_bounds` behave in the same manner as the earlier
+  `system_bounds`.
+
+  The following diagram illustrates the relationship between the exclusion and
+  inclusion bounds.
+  ```
+    inclusion.lower                              inclusion.upper
+  <-------|============|------------------|============|--------->
+                 exclusion.lower    exclusion.upper
+  ```
+  `----` values here are disallowed and wil be rejected, and
+  `====` values here are allowed and will be accepted.
+
 ## Bug Fixes
 
 <!-- Here goes notable bug fixes that are worth a special mention or explanation -->
--- a/proto/frequenz/api/microgrid/common.proto
+++ b/proto/frequenz/api/microgrid/common.proto
@@ -34,8 +34,61 @@ message Metric {
   // originates from.
   Bounds component_bounds = 3;
 
-  // The current bounds of the metric, as imposed by the overall system.
-  Bounds system_bounds = 4;
+  // These bounds indicate the range of values that are disallowed for the
+  // metric.
+  // If these bounds for a metric are [`lower`, `upper`], then this metric's
+  // `value` needs to comply with the constraints
+  // `value <= lower` OR `upper <= value`.
+  //
+  // It is important to note that these bounds work together with
+  // `system_inclusion_bounds`.
+  //
+  // E.g., for the system to accept a charge command,
+  // clients need to request power values within the bounds
+  // `[system_inclusion_bounds.lower, system_exclusion_bounds.lower]`.
+  // This means that clients can only request charge commands with values that
+  // are within the `system_inclusion_bounds`, but not within
+  // `system_exclusion_bounds`.
+  // Similarly, for the system to accept a discharge command,
+  // clients need to request power values within the bounds
+  // `[system_exclusion_bounds.upper, system_inclusion_bounds.upper]`.
+  //
+  // The following diagram illustrates the relationship between the bounds.
+  // ```
+  //   inclusion.lower                              inclusion.upper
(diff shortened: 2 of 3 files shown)


sahas-subramanian-frequenz:
Could you also mention somewhere, either in the PR or in the commit,  what the (generic) motivation for this change is, and how this change solves the problem?  Or linking/mentioning an issue is also fine.

What I understand is that the target power `v` needs to be between `system_bound_max.{lower, upper}`,  and `v` needs to be outside of `system_bounds_min.{lower, upper}`.

I guess that makes sense if the two bounds will always overlap, with `system_bounds_max` will be a broader range,  and we reject power requests within `system_bounds_min`, which I expect would lie around 0.  

So we are just trying to enforce a minimum power in either direction, by setting `system_bounds_min`.  

Is that accurate?  May be worth mention in the doc-comments.

tiyash-basu-frequenz:
> Could you also mention somewhere, either in the PR or in the commit, what the (generic) motivation for this change is, and how this change solves the problem? Or linking/mentioning an issue is also fine.

I have linked the issue.

tiyash-basu-frequenz:
> Is that accurate? May be worth mention in the doc-comments.

It's correct. The details of this behaviour is already present in the documentation, as well as in the release notes. Maybe you can point me toward exactly what detail is missing in the existing docs? I'll update those then.

sahas-subramanian-frequenz:
> Maybe you can point me toward exactly what detail is missing in the existing docs?

when I was reading the docs,  this was what took me the longest to figure out, that: the two bounds will always overlap, system_bounds_max will be a broader range, and we reject power requests within system_bounds_min, which I expect would lie around 0.

once I made that connection,  I could guess the point of having the two bounds.  So maybe just something like that - the relation between the min and the max bounds?

tiyash-basu-frequenz:
> the relation between the min and the max bounds?

I extended the existing docs to make it clearer how the min and the max bounds are supposed to be related. Please lemme know if the changes make sense.

sahas-subramanian-frequenz:
>      I extended the existing docs to make it clearer how the min and the max bounds are supposed to be related. Please lemme know if the changes make sense.

yup, I guess so.  Maybe @leandro-lucarella-frequenz wants to take a look as well?

leandro-lucarella-frequenz (commented):
Maybe because I'm missing the original context, but I find this extremely confusing and hard to understand.

If I finally got to understand it, after reading the docs on the specs, it seems like the `min` bounds is a range that will be rejected and `max` is a range that will be accepted. If this is the case, then it looks more like inclusion/exclusion lists, but with ranges. So if they are ranges, and using set algebra, then the values will be accepted only if in `max - min`, right?

```
-    max.lower                              max.upper   +
<-------|============|--------------|==========|-------->
                  min.lower      min.upper

---- rejected
==== accepted
```

Does this look OK? If so, what about naming these something more like `system_accept_range` and `system_reject_zone`? I would go with asymmetric names (i.e. not both zone or range) to try to make it more clear that one narrows the range of the other. It might be something else, but I would try to go more this route than min/max which is very confusing to me.

leandro-lucarella-frequenz:
> Could you also mention somewhere, either in the PR or in the commit, what the (generic) motivation for this change is, and how this change solves the problem? Or linking/mentioning an issue is also fine.

This was also a mystery for me, I think it would be good to mention the use case for this in the bounds docs, if not at least very briefly. I mean, even after understanding the concept, I'm not sure why the *reject zone* is needed, what would be the use case for it.

leandro-lucarella-frequenz:
> when I was reading the docs, this was what took me the longest to figure out, that: the two bounds will always overlap, system_bounds_max will be a broader range, and we reject power requests within system_bounds_min, which I expect would lie around 0.

Maybe also include the ascii diagram I put above for extra clarity.

tiyash-basu-frequenz:
> So if they are ranges, and using set algebra, then the values will be accepted only if in max - min, right?

Correct.

> what about naming these something more like `system_accept_range` and `system_reject_zone`?

How about `system_inclusion_bounds` and `system_exclusion_bounds`, to have something that's closer to existing set-theory definitions.

> even after understanding the concept, I'm not sure why the reject zone is needed, what would be the use case for it.

Sure, I can add something.

> Maybe also include the ascii diagram I put above for extra clarity.

Sounds like overkill, but sure.

leandro-lucarella-frequenz:
> How about `system_inclusion_bounds` and `system_exclusion_bounds`, to have something that's closer to existing set-theory definitions.

For me the bounds word here is confusing, even if at the end is represented by a `Bound`, for me this is a range. So `system_inclusion_range` and `system_exclusion_range` is better, but still, inclusion and exclusion IMHO is not very clear because it seems like a stretch to say a value is included or included, so IMHO `accept`/`reject` is much more accurate and clear. If you rather have both called `range`, is not so clear to me that using `range`/`zone` is better, so I would be fine with that too.

leandro-lucarella-frequenz:
> Sounds like overkill, but sure.

I agree it is redundant, but at least in my case, being a visual person, if I see the diagram I understand it instantly, while going through the text takes a bit more time and effort :)

leandro-lucarella-frequenz:
Just for the records, we talked with @tiyash-basu-frequenz and agreed on `system_inclusion_bounds` and `system_exclusion_bounds`. The rationale is:

* Inclusion and exclusion because seems to be a common term in maths and statistics to convey this meaning.
* Bounds because we want to prioritize uniformity of naming (components already use the term bounds) and the extra clarity can be added in the docs.

tiyash-basu-frequenz:
Updated. Ready for another look.

leandro-lucarella-frequenz (approved):
Haha, I see you made sure it is clarified everywhere :laughing: 

For me it would be also good to put the explanation in one place and in the other places say "see the documentation for XXX".

But anyway this is good to go for me! :shipit:

tiyash-basu-frequenz:
> For me it would be also good to put the explanation in one place and in the other places say "see the documentation for XXX".

I think it's better to keep it like this for now, given we are looking into extracting some common protobuf messages out into a new repo. This ensures that we do not need to copy documentation from different places, reducing chances of copy-paste errors.

I will merge the PR now.

PR #44, opened 2023-04-21
Removing all deprecated code

tiyash-basu-frequenz:
The following deprecated code has been removed:
* The `COMPONENT_CATEGORY_LOAD` variant from the `ComponentCategory` enum.
* The `COMPONENT_CATEGORY_JUNCTION` variant from the `ComponentCategory` enum.
* The RPCs `Charge` and `Discharge`, in favour of RPC `SetPowerActive`.
* The RPC `SetBounds`, in favour of RPCs `AddExclusiveBounds` and `AddInclusiveBounds`.

This removal also includes code that has been deprecated after the last major release. The reason to remove these deprecations now is to have a leaner API earlier, since we are already on the way for a major release now.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -85,6 +85,18 @@
 
   The RPC `SetBounds` has been deprecated.
 
+* [Removed deprecated code](https://github.com/frequenz-floss/frequenz-api-microgrid/pull/44)
+  The following deprecated code has been removed:
+  * The `COMPONENT_CATEGORY_LOAD` variant from the `ComponentCategory` enum.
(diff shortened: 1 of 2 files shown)


leandro-lucarella-frequenz (approved):
:fire:

leandro-lucarella-frequenz:
I would be good to have other eyes here, though, more experience with the historical use of the microgrid API, I'm just approving from the POV of the diff but I'm not fully aware of all the implications. Maybe @sahas-subramanian-frequenz should take a look too.

sahas-subramanian-frequenz (approved):
Yup,  that looks good.  The current SDK already doesn't use the LOAD and JUNCTION categories,  and we now a `SetPower` method and non-streaming methods for setting bounds,  all of which are great from the sdk point of view.

PR #50, opened 2023-05-05
Support solar and battery DC links for inverters

tiyash-basu-frequenz:
The `Inverter` message has been updated to support hybrid and solar inverters.

The `dc` field has been renamed to `dc_battery`, and a new field `dc_solar` has been added. The `dc_battery` field is used to report the DC electricity flowing to/from the battery, and the `dc_solar` field is used to report the DC electricity flowing to/from the solar panels.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,7 +6,15 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including if there are any deprecations and what they should be replaced with -->
+* [Updated inverters' DC links to support hybrid and solar inverters](https://github.com/frequenz-floss/frequenz-api-microgrid/pull/50)
+
+  The `Inverter` message has been updated to support hybrid and solar inverters.
+  The `dc` field has been renamed to `dc_battery`, and a new field `dc_solar`
+  has been added.
+  The `dc_battery` field is used to report the DC electricity flowing to/from
+  the linked battery, and is applicable to battery and hybrid inverters.
+  The `dc_solar` field is used to report the DC electricity flowing to/from the
+  linked solar panels, and is applicable to solar and hybrid inverters.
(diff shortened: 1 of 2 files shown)


tiyash-basu-frequenz:
FYI @thomas-nicolai-frequenz

thomas-nicolai-frequenz:
uh nice. lovely.

tiyash-basu-frequenz:
@leandro-lucarella-frequenz pinging you for a review.

leandro-lucarella-frequenz (commented):
LGTM. For someone with very little context about this like me it is not completely clear though. So now there could be inverters that are connected to both a PV and a battery?

So the upgrade path for users would be that sometimes now `dc_battery` should be used, and sometimes `dc_solar` (and sometimes both) based on the type? Looking at the `.proto` diff looks pretty clear, but maybe it should be stated more explicitly in the release notes too.

tiyash-basu-frequenz:
> So now there could be inverters that are connected to both a PV and a battery?

There are hybrid inverters in the market which can be connected to both a PV system and a battery system.

> So the upgrade path for users would be that sometimes now dc_battery should be used, and sometimes dc_solar (and sometimes both) based on the type? Looking at the .proto diff looks pretty clear, but maybe it should be stated more explicitly in the release notes too.

Done.

leandro-lucarella-frequenz (approved):

thomas-nicolai-frequenz:
> There are hybrid inverters in the market which can be connected to both a PV system and a battery system.

@leandro-lucarella-frequenz usually an inverter has one AC connection to the grid and one DC connection to the e.g. battery or solar system. In a hybrid inverter there is still one AC connection but two DC connections. This means a solar array can feed energy into a battery without the need to convert the energy from DC to AC and back from AC to DC. Thats the only major difference.

PR #51, opened 2023-05-22
Add enum variant for setting bounds on AC reactive power

tiyash-basu-frequenz:
This will allow clients to set bounds on a component's AC reactive power.


tiyash-basu-frequenz:
Rebased.

stefan-brus-frequenz (approved):

PR #53, opened 2023-05-22
Use the common frequenz API

tiyash-basu-frequenz:


leandro-lucarella-frequenz (commented):
LGTM. I would add a release notes entry mentioning the split, as downstream projects could be accessing some of the common objects or enums, and now they need to be imported from a different place in python, and if used explicitly, the `frequenz-api-common` pip package should be added to the dependencies too.

tiyash-basu-frequenz:
> I would add a release notes entry mentioning the split, as downstream projects could be accessing some of the common objects or enums, and now they need to be imported from a different place in python, and if used explicitly, the frequenz-api-common pip package should be added to the dependencies too.

Of course.
This is still a draft PR, and is WIP. I'll update people when it's ready for a review.

tiyash-basu-frequenz:
Closed this because the tests were waiting indefinitely for the outdated python versions.
I will create a new PR for this instead.

PR #57, opened 2023-05-23
Use proto definitions from the common API

tiyash-basu-frequenz:
The following proto definitions have been removed, and are being used from the `frequenz-api-common` repository instead:

* `ComponentCategory` -> `frequenz.api.common.components.ComponentCategory`
* `battery.Type` -> `frequenz.api.common.components.BatteryType`
* `common.Bounds` -> `frequenz.api.common.metrics.Bounds`
* `common.Metric` -> `frequenz.api.common.metrics.Metric`
* `common.Ac` -> `frequenz.api.common.metrics.electrical.Ac`
* `common.Dc` -> `frequenz.api.common.metrics.electrical.Dc`
* `ev_charger.Type` -> `frequenz.api.common.components.EVChargerType`
* `inverter.Type` -> `frequenz.api.common.components.InverterType`
* `sensor.Type` -> `frequenz.api.common.components.SensorType`

The pypi package `frequenz-api-common` is being added as a dependency to the python package definition, instead of generating the proto definitions using `protoc`. This is required, otherwise each proto library depending on `frequenz-api-common` will generate its own python modules for `frequenz-api-common`, resulting in multiple definition of the common data structures.

Note that this forces the minimum required python version of the `frequenz-api-microgrid` package to be 3.11, as a transitive dependency inherited from the `frequenz-api-common` package.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,6 +39,33 @@ If you have any issues with these dependencies, please check the
 `pyproject.toml` file and try installing the exact supported versions instead.
 
 
+Upgrading dependencies
+======================
+
+If you want to update the dependency `frequenz-api-common`, then you need to
+* update the submodule `frequenz-api-common`
+* update the version of the `frequenz-api-common` package in `pyproject.toml`
+* update the version of the `frequenz-api-common` package in
+`minimum-requirements-ci.txt`
+
+The version of `frequenz-api-common` used in all the three places mentioned
+above should be the same.
+
+Here is an example of upgrading the `frequenz-api-common` dependency to version
+`v0.2.0`:
+```sh
+ver="0.2.0"
+
+cd submodules/frequenz-api-common
+git remote update
+git checkout v${ver}
+cd ../..
+
+sed s/"frequenz-api-common == [0-9]\.[0-9]\.[0-9]"/"frequenz-api-common == ${ver}"/g -i pyproject.toml
+
+sed s/"frequenz-api-common == .*"/"frequenz-api-common == ${ver}"/g -i minimum-requirements-ci.txt
+```
+
 Releasing
 =========
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,14 +20,11 @@ classifiers = [
    "License :: OSI Approved :: MIT License",
    "Programming Language :: Python :: 3",
    "Programming Language :: Python :: 3 :: Only",
-   "Programming Language :: Python :: 3.7",
-   "Programming Language :: Python :: 3.8",
-   "Programming Language :: Python :: 3.9",
-   "Programming Language :: Python :: 3.10",
    "Topic :: Software Development :: Libraries",
 ]
-requires-python = ">= 3.7, < 4"
+requires-python = ">= 3.11, < 4"
 dependencies = [
+    "frequenz-api-common == 0.2.0",
     "googleapis-common-protos >= 1.56.2, < 2",
     "grpcio >= 1.47.0, < 2",
 ]
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,10 +53,6 @@ jobs:
         os:
           - ubuntu-20.04
         python-version:
-          - "3.7"
-          - "3.8"
-          - "3.9"
-          - "3.10"
           - "3.11"
     runs-on: ${{ matrix.os }}
     steps:
@@ -89,7 +85,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.10"
+          python-version: "3.11"
 
       - name: Install build dependencies
         run: |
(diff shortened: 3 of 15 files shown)


pyproject.toml
```diff
@@ -26,10 +27,9 @@ classifiers = [
    "Programming Language :: Python :: 3.10",
```
leandro-lucarella-frequenz: This looks like it should be removed. We are leaving only the `Python :: 3 :: Only` classifier in other projects as the range of supported versions is already specified in `requires-python`, so we avoid the extra maintenance cost of keeping classifiers up to date.

CONTRIBUTING.md
```diff
@@ -39,6 +39,33 @@ If you have any issues with these dependencies, please check the
...
+
+cd submodules/frequenz-api-common
+git remote update
+git checkout v${ver}
+cd ../..
+
+sed s/"frequenz-api-common == [0-9]\.[0-9]\.[0-9]"/"frequenz-api-common == ${ver}"/g -i pyproject.toml
+
+sed s/"frequenz-api-common == .*"/"frequenz-api-common == ${ver}"/g -i minimum-requirements-ci.txt
+```
```
leandro-lucarella-frequenz, line 67: A long time ago in a galaxy far, far away a tool was added to do this automatically... (https://github.com/frequenz-floss/frequenz-repo-config-python/issues/41)

pyproject.toml
```diff
@@ -1,5 +1,6 @@
 [build-system]
 requires = [
+    "frequenz-api-common == 0.2.0",
```
leandro-lucarella-frequenz: I don't think this is required as a build dependency. For building we are using the submodule.

pyproject.toml
```diff
@@ -26,10 +27,9 @@ classifiers = [
    "Programming Language :: Python :: 3.10",
    "Topic :: Software Development :: Libraries",
 ]
-requires-python = ">= 3.7, < 4"
+requires-python = ">= 3.11, < 4"
 dependencies = [
-    "googleapis-common-protos >= 1.56.2, < 2",
-    "grpcio >= 1.47.0, < 2",
```
leandro-lucarella-frequenz: This should be kept here, we need it in the python users to make sure the dependencies are fulfilled. Our (generated) code import stuff from the `googleapis-common-protos` python package.

```console
$ rg import.*google .direnv/python-3.11.2/lib64/python3.11/site-packages/frequenz/api/microgrid/ 
.direnv/python-3.11.2/lib64/python3.11/site-packages/frequenz/api/microgrid/microgrid_pb2_grpc.pyi
15:import google.protobuf.empty_pb2
16:import google.protobuf.wrappers_pb2

.direnv/python-3.11.2/lib64/python3.11/site-packages/frequenz/api/microgrid/microgrid_pb2.pyi
20:import google.protobuf.descriptor
21:import google.protobuf.internal.containers
```

Otherwise users will need to declare these dependencies themselves in their projects even if they don't use the google stuff in their code.

We also need to leave `grpcio`, stuff from the package `grpc` (provided by `grpcio` is used in the generated code:

```console
$ rg grpc .direnv/python-3.11.2/lib64/python3.11/site-packages/frequenz/api/microgrid/
.direnv/python-3.11.2/lib64/python3.11/site-packages/frequenz/api/microgrid/common_pb2_grpc.py
3:import grpc

.direnv/python-3.11.2/lib64/python3.11/site-packages/frequenz/api/microgrid/microgrid_pb2_grpc.pyi
17:import grpc
20:    def __init__(self, channel: grpc.Channel) -> None: ...
...
```

leandro-lucarella-frequenz (changes requested):
There are a few things that need to be fixed. I would also include an entry in the `RELEASE_NOTES` mentioning the new minimum supported Python version.

pyproject.toml
tiyash-basu-frequenz: Done :rocket: 
Should we keep the line `"Programming Language :: Python :: 3",`?

CONTRIBUTING.md
tiyash-basu-frequenz: I'd suggest still keeping it here to emphasize the point, and make it more self-contained.

pyproject.toml
tiyash-basu-frequenz: The tests do pass without this line. Removed. :rocket:

pyproject.toml
tiyash-basu-frequenz: Reverted :rocket:

tiyash-basu-frequenz:
> I would also include an entry in the RELEASE_NOTES mentioning the new minimum supported Python version.

Done.

pyproject.toml
leandro-lucarella-frequenz: Yeah, I would keep it as it is zero maintenance effort and it works in Python 3.

CONTRIBUTING.md
leandro-lucarella-frequenz: Yeah, it was a bad joke, is something that is probably nice to have in a future tool, but this is a bit far in the future... You can ignore.

tiyash-basu-frequenz: > Yeah, it was a bad joke

It was an awesome joke :D I failed to process it wisely though

leandro-lucarella-frequenz (approved):
:shipit:

PR #60, opened 2023-05-26
Add RPC to set reactive power

tiyash-basu-frequenz:
A new RPC, named `SetPowerReactive` has been added to set reactive power for inverters, and other components that support it. Also, the parameters to the RPC can be sent using the message `SetPowerReactiveParam`.


stefan-brus-frequenz (approved):
LGTM

PR #62, opened 2023-05-29
Clear release notes and update contribution docs

tiyash-basu-frequenz:


stefan-brus-frequenz (approved):

CONTRIBUTING.md
```diff
@@ -84,7 +84,7 @@ These are the steps to create a new release:
    for example:
 
    ```sh
-   git tag -s -F RELEASE_NOTES.md v0.0.1
+   git tag -s --cleanup=whitespace -F RELEASE_NOTES.md v0.0.1
```
leandro-lucarella-frequenz, line 87: This is strange, it shouldn't be a problem unless you are editing the file. From `git commit --help`:

>            default
>               Same as strip if the message is to be edited. Otherwise whitespace.

I have locally configured git to use the scissors method (`git config commit.cleanup scissors`), but of course we shouldn't depend on user configuration.

But it is true that at least it should have a mention that comments can be a problem if you edit the commit message, and also hardcoding a method should be safer too.

PR #102, opened 2023-08-09
Rename RPC StreamComponentData to SubscribeComponentData

### LINKED ISSUE #84: Rename RPCs named as `StreamX` to `GetXStream`
### What's needed?

`stream` as a verb refers to the act of publishing something. Therefore RPC names following the naming scheme `StreamX`, e.g., [`StreamComponentData`](https://github.com/frequenz-floss/frequenz-api-microgrid/blob/1abca32b1d0c62b5330aeecf1f38634f7c79a417/proto/frequenz/api/microgrid/microgrid.proto#L84), are ambigous. E.g., a client invoking `stream_component_data` could also indicate that the client intends to publish data, instead of consuming data from a stream.

### Proposed solution

A more appropriate name would be `GetXStream`. E.g., the RPC `StreamComponentData` could be renamed to `GetComponentDataStream`, removing any ambiguity in the intention of the method.

Edit:
The current consensus is to have `PublishX` methods to stream data, and `SubscribeX` methods to consume data from streams.

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
Using `stream` as a verb can be ambigous. It could mean publishing and subscribing to a data stream. `Subscribe` is a more unambiguous option.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -15,6 +15,9 @@
   `frequenz.api.common.location.Location` is being used instead. The `Location`
   message from the common API also has a `country_code` member.
 
+- The following gRPC method has been renamed:
+  `StreamComponentData` -> `SubscribeComponentData`
+
 ## New Features
 
 <!-- Here goes the main new features and examples or instructions on how to use them -->
--- a/proto/frequenz/api/microgrid/microgrid.proto
+++ b/proto/frequenz/api/microgrid/microgrid.proto
@@ -82,7 +82,7 @@ service Microgrid {
   }
 
   // Returns a stream containing data from a component with a given ID.
-  rpc StreamComponentData(ComponentIdParam) returns (stream ComponentData) {
+  rpc SubscribeComponentData(ComponentIdParam) returns (stream ComponentData) {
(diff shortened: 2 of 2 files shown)


RELEASE_NOTES.md
```diff
@@ -15,6 +15,9 @@
   `frequenz.api.common.location.Location` is being used instead. The `Location`
   message from the common API also has a `country_code` member.
 
+- The following gRPC methods have been renamed:
+  `StreamComponentData` -> `SubscribeComponentData`
```
stefan-brus-frequenz, line 19: Looks like a number of methods have been renamed, should they also be mentioned here?

tiyash-basu-frequenz: It was a mistake, fixed.

tiyash-basu-frequenz: Sorry, just realised the mistake. This PR is supposed to be renaming only the `StreamComponentData` method. The other methods will be renamed in different PR. I reverted the changes for those methods.

TalweSingh (approved):
Yes, that is more concise. LGTM!

PR #103, opened 2023-08-10
Rename RPCs that act on Components

### LINKED ISSUE #101: Add `Component` to methods that act on components
### What's needed?

The gRPC methods specified in the provided documentation exhibit clarity in terms of which entity the method verb pertains to. For instance, methods like `GetMicrogridMetadata`, `ListComponents`, `ListConnections`, and `StreamComponentData` explicitly convey the application of their respective verbs to specific entities. For example, `ListComponents` promptly indicates that the verb 'List' operates on 'Components', ensuring precise comprehension.

However, this clarity is not consistently present across all methods. Some methods such as `Start`, `HotStandby`, `ColdStandby`, `Stop`, and `ErrorAck` lack this specification. As these actions pertain to individual components, it's advisable to incorporate the term `Component` into their names to ensure clarity.

Furthermore, while methods like `SetPowerActive`, `SetPowerReactive`, `AddExclusionBounds`, and `AddInclusionBounds` accurately denote the action they perform, they fall short in specifying the entity upon which the action is executed. It's recommended to include the term `Component` in the names of these methods to enhance precision in conveying the entity of the action.

By adhering to this naming convention, the gRPC methods will consistently provide clear insight into the entities and actions they concern, contributing to a more comprehensible and user-friendly API documentation.

### Proposed solution

The following method names require revision:

- `AddExclusionBounds` should be renamed to `AddComponentExclusionBounds`.
- `AddInclusionBounds` should be renamed to `AddComponentInclusionBounds`.
- `SetPowerActive` should be renamed to `SetComponentPowerActive`.
- `SetPowerReactive` should be renamed to `SetComponentPowerReactive`.
- `Start` should be renamed to `StartComponent`.
- `HotStandby` should be renamed to `HotStandbyComponent`.
- `ColdStandby` should be renamed to `ColdStandbyComponent`.
- `Stop` should be renamed to `StopComponent`.

By incorporating the term "Component" into these method names, the actions will be distinctly associated with the relevant entity, leading to greater clarity and coherence within the API documentation.

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
Previously, these RPCs did not specify that these act on Components, which was a critical feature of these RPCs. This change adds `Component` to these RPCs, making them easier to read.

Here are the details of the RPCs that got renamed:
  `AddExclusionBounds`  -> `AddComponentExclusionBounds`
  `AddInclusionBounds`  -> `AddComponentInclusionBounds`
  `SetPowerActive`      -> `SetComponentPowerActive`
  `SetPowerReactive`    -> `SetComponentPowerReactive`
  `Start`               -> `StartComponent`
  `HotStandby`          -> `HotStandbyComponent`
  `ColdStandby`         -> `ColdStandbyComponent`
  `Stop`                -> `StopComponent`
  `ErrorAck`            -> `AckComponentError`

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -15,8 +15,17 @@
   `frequenz.api.common.location.Location` is being used instead. The `Location`
   message from the common API also has a `country_code` member.
 
-- The following gRPC method has been renamed:
+- The following gRPC methods have been renamed:
   `StreamComponentData` -> `SubscribeComponentData`
+  `AddExclusionBounds`  -> `AddComponentExclusionBounds`
+  `AddInclusionBounds`  -> `AddComponentInclusionBounds`
+  `SetPowerActive`      -> `SetComponentPowerActive`
+  `SetPowerReactive`    -> `SetComponentPowerReactive`
+  `Start`               -> `StartComponent`
+  `HotStandby`          -> `HotStandbyComponent`
+  `ColdStandby`         -> `ColdStandbyComponent`
+  `Stop`                -> `StopComponent`
+  `ErrorAck`            -> `AckComponentError`
 
 ## New Features
 
--- a/proto/frequenz/api/microgrid/microgrid.proto
+++ b/proto/frequenz/api/microgrid/microgrid.proto
@@ -134,7 +134,8 @@ service Microgrid {
   // ```
   // ---- values here are disallowed and wil be rejected
   // ==== vales here are allowed and will be accepted
-  rpc AddExclusionBounds(SetBoundsParam) returns (google.protobuf.Timestamp);
+  rpc AddComponentExclusionBounds(SetBoundsParam)
+    returns (google.protobuf.Timestamp);
 
   // Adds inclusion bounds for a given metric of a given component, and returns
   // the UTC timestamp until which the given inclusion bounds will stay in
@@ -171,7 +172,8 @@ service Microgrid {
   // ```
   // ---- values here are disallowed and wil be rejected
   // ==== vales here are allowed and will be accepted
-  rpc AddInclusionBounds(SetBoundsParam) returns (google.protobuf.Timestamp);
+  rpc AddComponentInclusionBounds(SetBoundsParam)
+    returns (google.protobuf.Timestamp);
 
   // Sets the active power output of a component with a given ID, provided the
   // component supports it.
@@ -186,7 +188,8 @@ service Microgrid {
   //
   // * Inverter: Sends the discharge command to the inverter, making it deliver
   //  AC power.
-  rpc SetPowerActive(SetPowerActiveParam) returns (google.protobuf.Empty) {
+  rpc SetComponentPowerActive(SetPowerActiveParam)
+    returns (google.protobuf.Empty) {
     option (google.api.http) = {
       get : "/v1/components/{component_id}/setPowerActive/{power}"
     };
@@ -199,7 +202,8 @@ service Microgrid {
   // E.g., an inverter may have a resolution of 88 VAr.
   // In such cases, the magnitude of power will be floored to the nearest
   // multiple of the resolution.
-  rpc SetPowerReactive(SetPowerReactiveParam) returns (google.protobuf.Empty) {
+  rpc SetComponentPowerReactive(SetPowerReactiveParam)
+    returns (google.protobuf.Empty) {
     option (google.api.http) = {
       get : "/v1/components/{component_id}/setPowerReactive/{power}"
     };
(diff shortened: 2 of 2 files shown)


matthias-wende-frequenz:
What about `ErrorAck`?

tiyash-basu-frequenz:
> What about `ErrorAck`?

Was missing in the docs. Thanks for pointing out. Added.

matthias-wende-frequenz (dismissed):

RELEASE_NOTES.md
```diff
@@ -15,8 +15,17 @@
...
   `StreamComponentData` -> `SubscribeComponentData`
+  `AddExclusionBounds`  -> `AddComponentExclusionBounds`
+  `AddInclusionBounds`  -> `AddComponentInclusionBounds`
+  `SetPowerActive`      -> `SetComponentPowerActive`
+  `SetPowerReactive`    -> `SetComponentPowerReactive`
+  `Start`               -> `StartComponent`
+  `HotStandby`          -> `HotStandbyComponent`
+  `ColdStandby`         -> `ColdStandbyComponent`
+  `Stop`                -> `StopComponent`
+  `ErrorAck`            -> `ErrorAckComponent`
```
llucax: Would it make more sense to name this `AckComponentError` while we are at it, as the verb is to acknowledge and what we are acknowledging is a component error, right?

tiyash-basu-frequenz: True, makes sense.

tiyash-basu-frequenz: Done. Also updated docs, commit, and PR description.

proto/frequenz/api/microgrid/microgrid.proto
```diff
@@ -265,7 +269,7 @@ service Microgrid {
   //
   // If any of the above mentioned actions for a given component has already
   // been performed, then this method call efffectively skips that action.
-  rpc ColdStandby(ComponentIdParam) returns (google.protobuf.Empty) {
+  rpc ColdStandbyComponent(ComponentIdParam) returns (google.protobuf.Empty) {
```
TalweSingh, line 272: Not sure where ColdStandby comes from.
If the Cold refers to the Component, maybe ColdComponentStandby, otherwise disregard this comment

TalweSingh (approved):

proto/frequenz/api/microgrid/microgrid.proto
tiyash-basu-frequenz: Good question!
The verb here is `ColdStandby`. For inverters it means that AC relays get opened. Inverters take about a minute to become operation from this state. This differs from `HotStandby`, where we put inverters in a state where we can make them operation again in a few milliseconds.

matthias-wende-frequenz: I'd vote for `{Hot, Cold}ComponentStandy`. It's the `ComponentStandby` RPC and we distinguish between the Hot and Cold variants.

tiyash-basu-frequenz: No sure about that. The verb here is `{Hot,Cold}Standby`, and we are following the verb_entity pattern. `{Hot,cold}ComponentStandby` breaks that pattern.

tiyash-basu-frequenz:
I will wait for 10 more minutes for people to look into the changes [here](https://github.com/frequenz-floss/frequenz-api-microgrid/compare/0d29dc35918f11b08a13a308f6d7babe8cf837c5..8012162204fbc13c3fe7a5dd9952710ea6328c66). If there are no further comments, will merge.

proto/frequenz/api/microgrid/microgrid.proto
llucax: I agree that cold-standby and hot-standby are too strong of a term to split. Another alternative would be not to use it as a verb and use something like `SetComponentColdStandby`.

tiyash-basu-frequenz: That would mean the other RPCs become `SetComponentStart`, `SetComponentStop`, etc.

~Another alternative could be `SetComponentStateX`. But I would go for it only if you really really dislike `HotStandby` and `ColdStandby` :)~

Note that `standby` is a legit verb to use here. Maybe we get rid of both the commands and just keep `Standby`? Then we have 3 clean verbs `Start`, `Standby`, and `Stop`? `HotStandby` get renamed to `Standby`, and instead of `ColdStandby` we use `Stop`? We do not really need a semantic distinction between `ColdStandby` and `Stop` in the Microgrid API, and it would make the API cleaner (I had been thinking about this for the v1 milestone)

llucax: If we really don't need the distinction, I'm all for simplifying APIs.

So `cold standby == stop` and `hot standby == standby`?

Another option could be to have a boolean argument for standby that says if it's hot or cold.

tiyash-basu-frequenz: > So cold standby == stop and hot standby == standby?

`coldStandby` gets removed. We don't need it here. Going to `stop` is better.
Indeed  `hotStandby == standby`

>Another option could be to have a boolean argument for standby that says if it's hot or cold.

In general, actions that move components from one state to another should have dedicated methods. Using parameters to do it leads to mistakes. However, in this case, we just remove `coldStandby`.

PR #145, opened 2023-09-29
Remove sensors from component category and represent the separately

### LINKED ISSUE #83: Represent sensors as as their own category
### What's needed?

Sensors will soon be represented as non-component entities, and will exist as a separate category. Look here for details: https://github.com/frequenz-floss/frequenz-api-common/issues/51.

The Microgrid API needs to be adjusted accordingly.

### Proposed solution

Create a new RPC named `StreamSensorData`, similar to [`StreamComponentData`](https://github.com/frequenz-floss/frequenz-api-microgrid/blob/1abca32b1d0c62b5330aeecf1f38634f7c79a417/proto/frequenz/api/microgrid/microgrid.proto#L84), that returns a stream of objects of the [`sensor.Sensor`](https://github.com/frequenz-floss/frequenz-api-microgrid/blob/1abca32b1d0c62b5330aeecf1f38634f7c79a417/proto/frequenz/api/microgrid/sensor.proto#L121) type.

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
This update introduces a dedicated RPC method for listing sensors in the microgrid, separating them from the "component" category. This new method allows for optional sensor filtering by ID or category. The change clarifies the term "component" to exclusively refer to electrical equipment like inverters and batteries, while sensors are now treated as a distinct entity.

### DIFF
--- a/proto/frequenz/api/microgrid/microgrid.proto
+++ b/proto/frequenz/api/microgrid/microgrid.proto
@@ -24,6 +24,7 @@ import "frequenz/api/microgrid/voltage_transformer.proto";
 import "frequenz/api/common/v1/components.proto";
 import "frequenz/api/common/v1/location.proto";
 import "frequenz/api/common/v1/metrics.proto";
+import "frequenz/api/common/v1/sensors.proto";
 
 import "google/api/annotations.proto";
 import "google/protobuf/empty.proto";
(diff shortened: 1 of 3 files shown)


thomas-nicolai-frequenz (dismissed):
All good from my side. We should however, clarify that when we say components we mean electrical components!

tiyash-basu-frequenz:
>We should however, clarify that when we say components we mean electrical components!

@thomas-nicolai-frequenz I added another commit to re-inforce components as electrical components. Ready for another look.

thomas-nicolai-frequenz (dismissed):

proto/frequenz/api/microgrid/microgrid.proto
```diff
@@ -490,3 +512,61 @@ message Connection {
...
+  uint64 component_id = 1;
+
+  // The target metric whose bounds have to be set.
+  TargetMetric target_metric = 2;
+
+  // The bounds for the target metric.
+  frequenz.api.common.v1.metrics.Bounds bounds = 3;
+}
+
+// Encapsulation of a component ID, intended to be used as a parameter for rpc
```
stefan-brus-frequenz, line 566: There are plenty of new `component_id` fields introduced here with the type `uint64`. Should these be `ComponentIDParam`? If not, I think the docs here (and for sensor below) should be clarified that these are only intended to be used as top-level parameters for the rpc methods themselves, and perhaps a little blurb explaining why.

stefan-brus-frequenz (commented):
Looks pretty good overall, with one reservation

proto/frequenz/api/microgrid/microgrid.proto
tiyash-basu-frequenz: This will go away once https://github.com/frequenz-floss/frequenz-api-microgrid/issues/149 is done. Is it okay if I handle this directly there?

stefan-brus-frequenz (approved):

PR #151, opened 2023-10-05
Add major version to package names

### LINKED ISSUE #77: Add major version to the protobuf package
### What's needed?

According to https://cloud.google.com/apis/design/versioning, `All Google API interfaces must provide a major version number, which is encoded at the end of the protobuf package`. For an example, look [here](https://github.com/googleapis/googleapis/blob/master/google/bigtable/v2/bigtable.proto#L17). Also, note how this affects the directory structure there (files in version 2 go into a directory named `v2`).

We do not do this ATM, as is evident [here](https://github.com/frequenz-floss/frequenz-api-microgrid/blob/v0.x.x/proto/frequenz/api/microgrid/microgrid.proto#L11). We need to do this before we make the first stable release (v1.0.0).

### Proposed solution

As mentioned here, look [here](https://github.com/googleapis/googleapis/blob/master/google/bigtable/v2/bigtable.proto#L17) for an example.

### Use cases

This achieves two objectives:
1. services, in theory, could start supporting multiple major versions, to support clients while they transition.
2. adopt it as a standard in the organization, regardless of whether point 1 is valid for some services or not.

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
The package names have been changed from
`frequenz.api.microgrid.<package>` to
`frequenz.api.microgrid.v1.<package>`. `v1` is the API's major version, and will be incremented for breaking changes.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -57,6 +57,10 @@
 - The RPC parameters have been renamed to be more consistent with the RPC names,
   and with each other.
 
+- The package names have been changed from `frequenz.api.microgrid.<package>` to
+  `frequenz.api.microgrid.v1.<package>`. `v1` is the API's major version, and
+  will be incremented for breaking changes.
+
 ## New Features
 
 <!-- Here goes the main new features and examples or instructions on how to use them -->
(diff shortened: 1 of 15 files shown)


llucax (dismissed):
Ha! I was writing that you missed the `py/` stuff, but I see you just updated it!

llucax (approved):

PR #153, opened 2023-10-05
Change RPCs

### LINKED ISSUE #152: Replace `HotStandby` and `ColdStandby` with `Standby`
### What's needed?

We do not need to distinguish between the hot and cold standby commands. Currently, they are used only for inverters, and the `HotStandby` command for most inverters corresponds to just sending a `discharge(0)` command (internally). So we could simplify the API by removing the `HotStandby` command, and renaming the `ColdStandby` command to `Standby`.

The `Standby` command should serve 2 functions:
1. It should bring the component from an operational state to a state where it consumes lower-than-usual energy.
2. It could take some time for a component to move to an operational state, once this command is issued.

### Proposed solution

_No response_

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
This PR includes the following changes:

* Remove RPC `HotStandby`
* Rename RPC `ColdStandby` to `Standby`
* Rename RPC `SubscribeComponentData` to `ReceiveComponentDataStream`
* Rename RPC `SubscribeSensorData` to `ReceiveSensorDataStream`

### DIFF
--- a/proto/frequenz/api/microgrid/v1/microgrid.proto
+++ b/proto/frequenz/api/microgrid/v1/microgrid.proto
@@ -289,27 +289,6 @@ service Microgrid {
     };
   }
 
-  // Sets the given component into a hot-standby state, from which it can return
-  // into an operational state within at most 5 seconds.
-  //
-  // Performs the following sequence actions for the following component
-  // categories:
-  //
-  // * Inverter: Checks if AC and DC relays are closed, then
-  //  * sets power to 0
-  //
-  // If any of the checks mentioned above fails, then the method call returns an
-  // error.
-  //
-  // If any of the above mentioned actions for a given component has already
-  // been performed, then this method call effectively skips that action.
-  rpc HotStandbyComponent(HotStandbyComponentRequest)
-    returns (google.protobuf.Empty) {
-    option (google.api.http) = {
-      get : "/v1/components/{id}/hotStandby"
-    };
-  }
-
   // Sets the given component into a standby state, from which it can take a
   // few minutes to return to an operational state. A transition to an
   // operational state can be triggered by calling the `StartComponent` RPC, or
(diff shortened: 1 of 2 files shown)


proto/frequenz/api/microgrid/v1/microgrid.proto
```diff
@@ -325,10 +304,10 @@ service Microgrid {
   //
   // If any of the above mentioned actions for a given component has already
   // been performed, then this method call efffectively skips that action.
-  rpc ColdStandbyComponent(ColdStandbyComponentRequest)
+  rpc StandbyComponent(StandbyComponentRequest)
```
thomas-nicolai-frequenz: This sounds strange. We are putting a component into standby. So maybe something `EnableComponentStandby` or something similiar? Here are some suggestions: 

* ToggleComponentStandbyMode
* SetComponentToStandby
* EnableStandbyForComponent
* SwitchToStandbyMode
* PutComponentInStandby

* ActivateStandbyMode / DeactivateStandbyMode
* ActivateComponentStandbyMode / DeactivateComponentStandbyMode
* ToggleStandbyState
* ChangeToStandbyMode

My question would be if a user puts an inverter into standby how can he activate it again? I think this we might also want to make clear in the documentation.

thomas-nicolai-frequenz: My preference would be to use a verb at the front. `SetComponentToStandby` or `ActivateComponentStandbyMode` would be my favourites.

tiyash-basu-frequenz: `PutComponentInStandby` sounds good. I have updated the new RPC name to `PutComponentInStandby`.

>My question would be if a user puts an inverter into standby how can he activate it again? I think this we might also want to make clear in the documentation.

Good call. I added it to the RPC docs.

stefan-brus-frequenz (approved):

proto/frequenz/api/microgrid/v1/microgrid.proto
llucax: `StandbyComponent` also sounded a bit weird to me at first, but I think it had a nicer symmetry with `StartComponent` and `StopComponent`, IMHO now it is harder to see those 3 methods are related. As a final suggestion, I would put `PauseComponent` to the table, but I don't have any major complains if you prefer to keep `PutComponentInStandby`.

llucax: (just unresolved so my last comment is not lost)

Also, not sure if it would be worth clarifying in `SetComponentPowerActive` docs that this will start the component if it was in standby/paused.

PR #158, opened 2023-10-10
Bump repo-config to v0.6.2

TalweSingh:


pytests/test_microgrid.py
```diff
@@ -3,22 +3,23 @@
 
 """Tests for the frequenz.api.microgrid package."""
 
-# pylint: disable=import-outside-toplevel
```
llucax, line 6: Do you know why did you need to move this to every function? Or is it because it is like that in the repo-config template? If that is the case, I think we should update the repo-config template instead.

llucax (dismissed):

pytests/test_microgrid.py
TalweSingh: It's because of repo-config template, I can create an issue for this.
Also can you update the branch protection rules here as well pls

tiyash-basu-frequenz (approved):

PR #179, opened 2023-11-16
Add support for filtering metrics in data streams

### LINKED ISSUE #178: Add Filter message and add it to the `ReceiveComponentDataStreamRequest`
### What's needed?

The `ReceiveComponentDataStream` rpc returns a stream of responses each of which containing a list of `metric_samples` in the data. The only parameter that can be passed is the `component_id` which implies that the returned stream contains all metrics that are streamable by this components.

From a user perspective not all of these metrics are used and thus are ignored. Still the user has to iterate over the full list. 

### Proposed solution

Instead the users could provide a filter with only the metrics that they need.

```protobuf
message ReceiveSensorDataStreamRequest {
  // The sensor ID to subscribe to.
  uint64 sensor_id = 1;
  repeated metric filter 
}
```

### Use cases

Creating high level objects in the SDK that stream composed data but aren't interested in the full data stream. Among the metrics that might not be of interest are all kinds of harmonics, (re)active energy, power factor, total harmonic distortion.

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
This change adds support for filtering metrics in data streams. The request messages for receiving data streams have now been extended to consist of a list of metrics to be streamed. This allows the user to request only the metrics they are interested in, instead of receiving all of them. If this list is empty, then no data will be streamed, and the service will return an error.

### DIFF
--- a/proto/frequenz/api/microgrid/v1/microgrid.proto
+++ b/proto/frequenz/api/microgrid/v1/microgrid.proto
@@ -413,8 +413,30 @@ message ListConnectionsResponse {
 
 // Request parameters for the RPC `ReceiveComponentDataStream`.
 message ReceiveComponentDataStreamRequest {
+  // A message for specifying a filter to apply to the stream.
+  message ComponentDataStreamFilter {
+    // List of metrics to return. Only the specified metrics will be returned.
+    //
+    // !!! note
+    //     At least one metric must be specified. If no metric is specified,
+    //     then the stream will return an error.
+    //
+    // !!! note
+    //     Components may not support all metrics. If a component does not
+    //     support a given metric, then the returned data stream will not
+    //     contain that metric.
+    repeated frequenz.api.common.v1.metrics.Metric metrics = 1;
+  }
+
   // The component ID to subscribe to.
   uint64 component_id = 1;
+
+  // The filter to apply to the stream.
+  //
+  // This field is optional. If this is not provided, then the stream will
+  // return all metrics for the given component. If this is provided, then the
+  // stream will return only the metrics specified in the filter.
+  ComponentDataStreamFilter filter = 2;
 }
 
 // A data sample from a component in the microgrid.
@@ -426,8 +448,30 @@ message ReceiveComponentDataStreamResponse {
 
 // Request parameters for the RPC `ReceiveSensorDataStream`.
 message ReceiveSensorDataStreamRequest {
+  // A message for specifying a filter to apply to the stream.
+  message SensorDataStreamFilter {
+    // List of metrics to return. Only the specified metrics will be returned.
+    //
+    // !!! note
+    //     At least one metric must be specified. If no metric is specified,
+    //     then the stream will return an error.
+    //
+    // !!! note
+    //     Sensors may not support all metrics. If a sensor does not support a
+    //     given metric, then the returned data stream will not contain that
+    //     metric.
+    repeated frequenz.api.common.v1.metrics.Metric metrics = 1;
+  }
+
   // The sensor ID to subscribe to.
   uint64 sensor_id = 1;
+
+  // The filter to apply to the stream.
+  //
+  // This field is optional. If this is not provided, then the stream will
+  // return all metrics for the given component. If this is provided, then the
+  // stream will return only the metrics specified in the filter.
+  SensorDataStreamFilter filter = 2;
 }
 
 // A data sample from a sensor in the microgrid.
--- a/proto/frequenz/api/microgrid/v1/microgrid.proto
+++ b/proto/frequenz/api/microgrid/v1/microgrid.proto
@@ -1,3 +1,5 @@
+// protolint:disable MAX_LINE_LENGTH
+
 // Frequenz Microgrid API
 //
 // Copyright:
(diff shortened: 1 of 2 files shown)


tiyash-basu-frequenz:
As discussed, the filters are optional now, so that users have a convenient way of fetching all metrics from a component, given that the list of available metrics can vary between components.

tiyash-basu-frequenz:
I updated the PR based on comments [here](https://github.com/frequenz-floss/frequenz-api-microgrid/commit/1479f5741e589055e3075ec28db8dbabf3077a3d#r132817966).

I also changed the filter by moving the `component_id` out of it. The request returning all data for all components by default feels like overkill. Lemme know if this makes sense, and if I should move `component_id` back into the filter message.

proto/frequenz/api/microgrid/v1/microgrid.proto
```diff
@@ -413,15 +413,25 @@ message ListConnectionsResponse {
 
 // Request parameters for the RPC `ReceiveComponentDataStream`.
 message ReceiveComponentDataStreamRequest {
+  // A message for specifying a filter to apply to the stream.
+  message ComponentDataStreamFilter {
+    // List of metrics to return. Only the specified metrics will be returned.
+    // A complete list of metrics can be found [here](https://github.com/frequenz-floss/frequenz-api-common/blob/v0.5.0/proto/frequenz/api/common/v1/metrics/metric_sample.proto#L65-L66).
+    //
+    // !!! note
+    //     At least one metric must be specified. Failure to do so will result in an empty response.
```
TalweSingh: In the commit is says the RPC will return an error. Maybe also add this here in the docs.

proto/frequenz/api/microgrid/v1/microgrid.proto
```diff
@@ -413,15 +413,25 @@ message ListConnectionsResponse {
...
+    repeated frequenz.api.common.v1.metrics.Metric metrics = 1;
+  }
+
   // The component ID to subscribe to.
   uint64 component_id = 1;
 
-  // List of metrics to return. Only the specified metrics will be returned.
```
TalweSingh: These were just added in the last commit, right? I think I would squash them.

TalweSingh (commented):
2 optional comments. Otherwise LGTM!

proto/frequenz/api/microgrid/v1/microgrid.proto
tiyash-basu-frequenz: Good catch, thanks! Updated.

proto/frequenz/api/microgrid/v1/microgrid.proto
tiyash-basu-frequenz: Done.

TalweSingh (dismissed):

proto/frequenz/api/microgrid/v1/microgrid.proto
```diff
@@ -411,8 +413,26 @@ message ListConnectionsResponse {
 
 // Request parameters for the RPC `ReceiveComponentDataStream`.
 message ReceiveComponentDataStreamRequest {
+  // A message for specifying a filter to apply to the stream.
+  message ComponentDataStreamFilter {
+    // List of metrics to return. Only the specified metrics will be returned.
+    // A complete list of metrics can be found [here](https://github.com/frequenz-floss/frequenz-api-common/blob/v0.5.0/proto/frequenz/api/common/v1/metrics/metric_sample.proto#L65-L66).
```
thomas-nicolai-frequenz: I think it would be good to mention that not all metrics are supported by all components.

proto/frequenz/api/microgrid/v1/microgrid.proto
```diff
@@ -424,8 +444,26 @@ message ReceiveComponentDataStreamResponse {
 
 // Request parameters for the RPC `ReceiveSensorDataStream`.
 message ReceiveSensorDataStreamRequest {
+  // A message for specifying a filter to apply to the stream.
+  message SensorDataStreamFilter {
+    // List of metrics to return. Only the specified metrics will be returned.
+    // A complete list of metrics can be found [here](https://github.com/frequenz-floss/frequenz-api-common/blob/v0.5.0/proto/frequenz/api/common/v1/metrics/metric_sample.proto#L65-L66).
```
thomas-nicolai-frequenz: I would not put in the link here. The import path is already given and also line 455 gives the entire path where to find it. This is too much.

proto/frequenz/api/microgrid/v1/microgrid.proto
thomas-nicolai-frequenz: Same here, I suggest to remove the hyperlink. Feels too much. See down below.

tiyash-basu-frequenz: Both comments have been addressed now.

proto/frequenz/api/microgrid/v1/microgrid.proto
tiyash-basu-frequenz: Done.

stefan-brus-frequenz (approved):

PR #183, opened 2023-11-20
Return valid-until timestamps from set-power RPCs

### LINKED ISSUE #157: Return UTC timestamp until when requested power level is valid
### What's needed?

The commands `SetComponentPowerActive` and `SetComponentPowerReactive` do not return anything at the moment.

The commands should, at least, return the timestamp until which the requested power will be active unless no further command is sent. Otherwise clients cannot know if the service will issue the command for an indefinitely long duration, or will stop it after a definite amount of time.

### Proposed solution

The `SetComponentPowerActive` RPC should return an object of type `SetComponentPowerActiveResponse`. The `SetComponentPowerActiveResponse` message should be the following:

```proto
message SetComponentPowerActiveResponse {
    google.protobuf.Timestamp valid_until_ts = 1;
}
```

(Similar for the `SetComponentPowerReactive` RPC)

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
The RPCs to set power now return timestamps until which the command will stay in effect. The component's active/reactive power will be set to 0 after this timestamp, if the API receives no further requests to change it before then.

By default, this timestamp will be the current time plus 60 seconds.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -48,6 +48,12 @@
   If this list is empty, then no data will be streamed, and the service will
   return an error.
 
+- The RPC `SetComponentPowerActive`now returns a timestamp until which the
+  command will stay in effect. The component's active power will be set to 0
+  after this timestamp, if the API receives no further requests to change it
+  before then. By default, this timestamp will be the current time plus 60
+  seconds.
+
 ## New Features
 
 <!-- Here goes the main new features and examples or instructions on how to use them -->
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -54,6 +54,12 @@
   before then. By default, this timestamp will be the current time plus 60
   seconds.
 
+- The RPC `SetComponentPowerReactive`now returns a timestamp until which the
+  command will stay in effect. The component's reactive power will be set to 0
+  after this timestamp, if the API receives no further requests to change it
+  before then. By default, this timestamp will be the current time plus 60
+  seconds.
+
 ## New Features
 
 <!-- Here goes the main new features and examples or instructions on how to use them -->
--- a/proto/frequenz/api/microgrid/v1/microgrid.proto
+++ b/proto/frequenz/api/microgrid/v1/microgrid.proto
@@ -220,7 +220,15 @@ service Microgrid {
     returns (AddComponentInclusionBoundsResponse);
 
   // Sets the active power output of a component with a given ID, provided the
-  // component supports it.
+  // component supports it. The power output is specified in watts.
+  //
+  // The power output can be -ve or +ve, depending on whether the component is
+  // supposed to be discharging or charging, respectively.
+  //
+  // The return value is the timestamp until which the given power command will
+  // stay in effect. After this timestamp, the component's active power will be
+  // set to 0, if the API receives no further command to change it before then.
+  // By default, this timestamp will be set to the current time plus 60 seconds.
   //
   // Note that the target component may have a resolution of more than 1 W.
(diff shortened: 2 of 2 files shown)


RELEASE_NOTES.md
```diff
@@ -42,6 +42,16 @@
   As a result, all files besides `microgrid.proto` became obsolete, and
   therefore, have been removed.
 
+- The RPC `SetComponentPowerActive`now returns a timestamp until which the
+  command will stay in effect. The component's active power will be set to 0
+  after this timestamp, if the API receives no further requests to change it
+  before then. Typically, this timestamp will be the current time 60 seconds.
```
matthias-wende-frequenz: current time plus ...

RELEASE_NOTES.md
```diff
@@ -42,6 +42,16 @@
...
+
+- The RPC `SetComponentPowerReactive`now returns a timestamp until which the
+  command will stay in effect. The component's reactive power will be set to 0
+  after this timestamp, if the API receives no further requests to change it
+  before then. Typically, this timestamp will be the current time 60 seconds.
```
matthias-wende-frequenz: ditto

proto/frequenz/api/microgrid/v1/microgrid.proto
```diff
@@ -218,7 +218,15 @@ service Microgrid {
...
-  // component supports it.
+  // component supports it. The power output is specified in watts.
+  //
+  // The power output can be -ve or +ve, depending on whether the component is
+  // supposed to be discharging or charging, respectively.
+  //
+  // The return value is the timestamp until which the given power command will
+  // stay in effect. After this timestamp, the component's active power will be
+  // set to 0, if the API receives no further command to change it before then.
+  // Typically, this timestamp will be set to the current time plus 60 seconds.
```
matthias-wende-frequenz: Typically is not a very strong definition. I'd suggest to refer to a default value if it exists or write that this is implementation specific but typically (or usually), ...

proto/frequenz/api/microgrid/v1/microgrid.proto
```diff
@@ -231,21 +239,30 @@ service Microgrid {
...
+  // component supports it. The power output is specified in VAr.
+  //
+  // The power output can be -ve or +ve, depending on whether the component is
+  // supposed to be delivering inductive or capacitive power, respectively.
+  //
+  // The return value is the timestamp until which the given power command will
+  // stay in effect. After this timestamp, the component's reactive power will
+  // be set to 0, if the API receives no further command to change it before
+  // then.
+  // Typically, this timestamp will be set to the current time plus 60 seconds.
```
matthias-wende-frequenz: Same comment about typically applies here.

matthias-wende-frequenz (commented):
Looks good. I've only some minor comments about the documentation.

proto/frequenz/api/microgrid/v1/microgrid.proto
tiyash-basu-frequenz: I changed it to "By default". This needs to be guaranteed in the implementation.

proto/frequenz/api/microgrid/v1/microgrid.proto
tiyash-basu-frequenz: I changed it to "By default". This needs to be guaranteed in the implementation.

matthias-wende-frequenz (dismissed):

tiyash-basu-frequenz:
Needs another approval after rebase.

llucax (approved):

PR #205, opened 2024-01-30
Bump the api-common dependency to v0.5.4

llucax:
The new v0.5.4 release of the api-common repository is now backwards compatible with the old v0.3.x releases. We bump the dependency so microgrid API can be used together with other APIs that depend on the newer common API version.

Please note that the googleapis-common-protos dependency is also bumped to v1.56.4, which is the version that api-common v0.5.4 depends on.

Strictly speaking, this is a breaking change too, as projects might need to bump their googleapis-common-protos dependency to v1.56.4 too if they specify it explicitly, but this is highly unlikely and very easy to fix.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,4 +2,8 @@
 
 ## Summary
 
-This release doesn't introduce any visible changes, it's done mainly to be able to build the new [documentation website](https://frequenz-floss.github.io/frequenz-api-microgrid/).
+This release bumps the `frequenz-api-common` dependency to v0.5.3, allowing downstream projects to use a newer `frequenz-api-common` version too.
+
+Please note that the googleapis-common-protos dependency is also bumped to v1.56.4, which is the version that api-common v0.5.4 depends on.
+    
+Strictly speaking, this is a breaking change, as you might need to bump your googleapis-common-protos dependency to v1.56.4 too if it is specified explicitly, but this is highly unlikely to happen and very easy to fix.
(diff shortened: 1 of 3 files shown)


llucax:
I'm not sure why tests are not being run here... :thinking:

leandro-lucarella-frequenz:
I guess it was because I made the PR originally against the wrong branch, now I force-pushed and they are running...

leandro-lucarella-frequenz:
This needs updating the dependency when common API v0.5.3 is released. The test with nox CI check also fails because we need a release of common API v0.5.3 first for it to pass (otherwise it can't parse the minimum version from the string).

The tests that are not running are just because we are in an old branch that used different check names.

llucax:
I guess it was because I made the PR originally against the wrong branch, now I force-pushed and they are running...

llucax:
This needs updating the dependency when common API v0.5.4 is released. The test with nox CI check also fails because we need a release of common API v0.5.4 first for it to pass (otherwise it can't parse the minimum version from the string).

The tests that are not running are just because we are in an old branch that used different check names.

llucax:
API common v0.5.4 was released, so this PR is updated too and ready for review (and merge).

llucax:
FYI, there are a couple of CI checks that are expected but will never run because they are not present in this branch, but all tests are passing. We'll need to force-merge once this is approved.

Marenz (approved):

tiyash-basu-frequenz (approved):

llucax:
I will force-merge this, as the tests will never complete.

PR #218, opened 2024-03-12
Support repo-config v0.9 properly

tiyash-basu-frequenz:
This repo recently moved to repo-config v0.9. While the dependabot PR for the upgrade passed, the release notes for repo-config suggest more changes. This commit tries to add these additional requirements for repo-config v0.9 upgrade.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -7,34 +7,44 @@
 # https://github.com/marketplace/actions/labeler
 
 "part:docs":
-  - "**/*.md"
-  - "docs/**"
-  - "examples/**"
-  - LICENSE
+  - changed-files:
+    - any-glob-to-any-file:
+      - "**/*.md"
+      - "docs/**"
+      - "examples/**"
+      - LICENSE
 
 "part:tests":
-  - "**/conftest.py"
-  - "pytests/**"
+  - changed-files:
+    - any-glob-to-any-file:
+      - "**/conftest.py"
+      - "pytests/**"
 
 "part:tooling":
-  - "**/*.ini"
-  - "**/*.toml"
-  - "**/*.yaml"
-  - "**/*.yml"
-  - "**/conftest.py"
-  - ".editorconfig"
-  - ".git*"
-  - ".git*/**"
-  - "docs/*.py"
-  - CODEOWNERS
-  - MANIFEST.in
-  - noxfile.py
+  - changed-files:
+    - any-glob-to-any-file:
+      - "**/*.ini"
+      - "**/*.toml"
+      - "**/*.yaml"
+      - "**/*.yml"
+      - "**/conftest.py"
+      - ".editorconfig"
+      - ".git*"
+      - ".git*/**"
+      - "docs/*.py"
+      - CODEOWNERS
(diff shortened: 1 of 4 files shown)


.github/labeler.yml
```diff
@@ -7,34 +7,44 @@
 # https://github.com/marketplace/actions/labeler
 
 "part:docs":
```
Marenz, line 9: While this change is good and important, it requires you to also use labeler ver. 5: https://github.com/frequenz-floss/frequenz-api-microgrid/blob/3ad22440c2039207de4e901ec465fe4ab756b04d/.github/workflows/labeler.yml#L21

Marenz: See also https://github.com/frequenz-floss/frequenz-api-microgrid/pull/190

Marenz: (you probably can just cherry-pick that commit from the PR)

Marenz (commented):
Without the labeler update this will break labeling after merge.
Also I notice you didn't add the fix to have git-support in pyproject for the test-installation ci job. 
Which isn't a problem as long as git-urls are not used as repository dependencies..
Here is the patch in question:

```
patch -p1 <<'EOF'
diff --git a/.github/containers/test-installation/Dockerfile b/.github/containers/test-installation/Dockerfile
index 772b2ae..2494545 100644
--- a/.github/containers/test-installation/Dockerfile
+++ b/.github/containers/test-installation/Dockerfile
@@ -6,7 +6,12 @@

 FROM --platform=${TARGETPLATFORM} python:3.11-slim

-RUN python -m pip install --upgrade --no-cache-dir pip
+RUN apt-get update -y && \
+    apt-get install --no-install-recommends -y \
+    git && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* && \
+    python -m pip install --upgrade --no-cache-dir pip

 COPY dist dist
 RUN pip install dist/*.whl && \
EOF
```

.github/labeler.yml
tiyash-basu-frequenz: Done :rocket:

tiyash-basu-frequenz:
>Also I notice you didn't add the fix to have git-support in pyproject for the test-installation ci job.

It is because there is no test-installation step for this repo (which is by intention), so this step does not apply here.

Marenz (approved):

thea-leake (approved):
LGTM!

.github/workflows/labeler.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         # only use hashes to pick the action to execute (instead of tags or branches).
         # For more details read:
         # https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
-        uses: actions/labeler@ac9175f8a1f3625fd0d4fb234536d26811351594  # 4.3.0
+        uses: actions/labeler@v5.0.0
```
llucax, line 21: We should use the hash here for security reasons.

llucax: * https://github.com/frequenz-floss/frequenz-api-microgrid/pull/230

PR #225, opened 2024-03-25
Cleanup

tiyash-basu-frequenz:


hannah-stevenson-frequenz (approved):

RELEASE_NOTES.md
```diff
@@ -14,4 +14,4 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+- Removed unused imports
```
Marenz, line 17: Is that really a _notable_ bug fix?

Marenz: Oh, and then it was removed again

Marenz (approved):
I mean, approve, sure, but maybe you want to not add the release notes message in the first place instead of removing it in the next commit ;)

Marenz:
Ah, too late I guess

RELEASE_NOTES.md
tiyash-basu-frequenz: Not sure what you mean. No release note is being removed in any commit here.

tiyash-basu-frequenz:
>I mean, approve, sure, but maybe you want to not add the release notes message in the first place instead of removing it in the next commit ;)

You are probably mistaken here (or perhaps I am :) ). Let's discuss in the thread https://github.com/frequenz-floss/frequenz-api-microgrid/pull/225#discussion_r1537909865

RELEASE_NOTES.md
tiyash-basu-frequenz: If you mean whether `Removed unused imports` is notable or not, then yes. Otherwise the gRPC gateway build fails.
A bit of a concern though, given that we did not catch this in the CI. Will definitely need to fix it.

Marenz: I am not sure what happened there, I saw different things at first. Looking good now regarding adding/removing.

Hmm maybe we should create an issue then to create a test for that specific part

tiyash-basu-frequenz: Good point. Done: https://github.com/frequenz-floss/frequenz-api-microgrid/issues/228

PR #229, opened 2024-03-26
Add a new RPC for adding component bounds

### LINKED ISSUE #180: Remove the `Inclusion`/`Exclusion` terminology in RPCs for adding bounds
### What's needed?

Our API processes `Bounds` with a conservative approach, allowing clients to specify them as either 'Inclusion' or 'Exclusion' types. Clients provide one pair of bounds at a time, tagged accordingly. The philosophy is to ensure safety by:

1. **Intersecting Inclusion Bounds**: Multiple Inclusion bounds intersect to form the narrowest possible range, focusing on restrictive aggregation.
2. **Unioning Exclusion Bounds**: Multiple Exclusion bounds merge into a broad range, aiming for extensive coverage.

This method ensures that the aggregated bounds is never less conservative than any of the input bounds.

However, in reporting bounds within data streams, we've moved away from labeling them as 'Inclusion' or 'Exclusion'. Instead, we return a single list of bounds, accepting any input request parameter falling within these bounds. This shift leads to a discrepancy: users specify bounds with labels, but receive an unlabeled, aggregated list. This mismatch could cause confusion about the bounds' interpretation.

### Proposed solution

A potential solution is to let users submit a list of bounds without specifying 'Inclusion' or 'Exclusion'. Our service could then intersect these lists to derive a final, conservatively aggregated set of bounds, aligning with our cautious approach.

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
Closes #180

### DIFF
--- a/proto/frequenz/api/microgrid/v1/microgrid.proto
+++ b/proto/frequenz/api/microgrid/v1/microgrid.proto
@@ -130,6 +130,38 @@ service Microgrid {
     };
   }
 
+  // Adds inclusion bounds for a given metric of a given component. Returns the
+  // time at which the bounds will expire (as a timestamp (UTC))
+  //
+  // Inclusion bounds give the range that the system will try to keep the
+  // metric within. If the metric goes outside of these bounds, the system will
+  // try to bring it back within the bounds.
+  // If the bounds for a metric are [[`lower_1`, `upper_1`],
+  // [`lower_2`, `upper_2`]], then this metric's `value` needs to comply with
+  // the constraints
+  // `lower_1 <= value <= upper_1` OR `lower_2 <= value <= upper_2`.
+  //
+  // If multiple inclusion bounds have been provided for a metric, then the
+  // overlapping bounds are merged into a single bound, and non-overlapping
+  // bounds are kept separate.
+  // E.g. if the bounds are [[0, 10], [5, 15], [20, 30]], then the resulting
+  // bounds will be [[0, 15], [20, 30]].
+  //
+  // The following diagram illustrates how bounds are applied:
+  // ```
+  //   lower_1  upper_1
+  // <----|========|--------|========|-------->
+  //                     lower_2  upper_2
+  // ```
+  // The bounds in this example are `[[lower_1, upper_1], [lower_2, upper_2]]`.
+  // ---- values here are considered out of range.
+  // ==== values here are considered within range.
+  //
+  // Note that for power metrics, regardless of the bounds, 0W is always
+  // allowed.
+  rpc AddComponentBounds(AddComponentBoundsRequest)
+    returns (AddComponentBoundsResponse);
+
   // Adds exclusion bounds for a given metric of a given component, and returns
   // the UTC timestamp until which the given exclusion bounds will stay in
   // effect.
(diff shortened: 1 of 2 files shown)


thomas-nicolai-frequenz (dismissed):
LGTM

stefan-brus-frequenz (dismissed):
LGTM

proto/frequenz/api/microgrid/v1/microgrid.proto
```diff
@@ -130,92 +130,36 @@ service Microgrid {
...
-  // `[system_exclusion_bounds.upper, system_inclusion_bounds.upper]`.
-  //
-  // The following diagram illustrates the relationship between the bounds.
+  // metric. If these bounds for a metric are [[`lower_1`, `upper_1`],
+  // [`lower_2`, `upper_2`]], then
+  // this metric's `value` needs to comply with the constraint
+  // `lower_1 <= value <= upper_1` AND `lower_2 <= value <= upper_2`.
+  //
+  // If multiple inclusion bounds have been provided for a metric, then the
+  // overlapping bounds are coalesced into a single bound, and disjointed
```
hannah-stevenson-frequenz:
```suggestion
  // overlapping bounds are merged into a single bound, and non-overlapping
```

proto/frequenz/api/microgrid/v1/microgrid.proto
```diff
@@ -130,6 +130,37 @@ service Microgrid {
     };
   }
 
+  // Adds inclusion bounds for a given metric of a given component, and returns
+  // the UTC timestamp until which the given inclusion bounds will stay in
+  // effect.
```
hannah-stevenson-frequenz: I'm not sure its wrong but maybe it would read better as:

```
Adds inclusion bounds for a given metric of a given component. Returns the time at which the bounds will expire (as a timestamp (UTC)) 
```

proto/frequenz/api/microgrid/v1/microgrid.proto
```diff
@@ -130,6 +130,37 @@ service Microgrid {
...
+  //
+  // Inclusion bounds refer to the range of values that are allowed for the
```
hannah-stevenson-frequenz: a given metric may be set to?

I think we are abusing the word metric here...if we are including things like power to discharge at its not really a metric. A metric is something that you measure - if you set it it sounds more like a property.

proto/frequenz/api/microgrid/v1/microgrid.proto
```diff
@@ -130,6 +130,37 @@ service Microgrid {
...
+  // bounds are kept disjointed.
+  // E.g. if the bounds are [[0, 10], [5, 15], [20, 30]], then the resulting
+  // bounds will be [[0, 15], [20, 30]].
+  //
+  // The following diagram illustrates how bounds are applied:
+  // ```
+  //   lower_1  upper_1
+  // <----|========|--------|========|-------->
+  //                     lower_2  upper_2
+  // ```
```
hannah-stevenson-frequenz, line 155: this contradicts the above..."needs to comply with the constraint
  // `lower_1 <= value <= upper_1` AND `lower_2 <= value <= upper_2`". If the statement is true then no value should be allowed in this example.

proto/frequenz/api/microgrid/v1/microgrid.proto
```diff
@@ -507,6 +538,27 @@ enum ComponentBoundsTargetMetric {
...
+message AddComponentBoundsRequest {
+  // The ID of the target component.
+  uint64 component_id = 1;
+
+  // The target metric whose bounds have to be set.
+  ComponentBoundsTargetMetric target_metric = 2;
+
+  // A list of bounds for the target metric.
+  // While incorporating these bounds, the API will ensure that overlapping
+  // pairs of bounds are coalesced into a single pair of bounds, and disjointed
```
hannah-stevenson-frequenz: merged and non-overlapping might be better terms?

proto/frequenz/api/microgrid/v1/microgrid.proto
tiyash-basu-frequenz: Done

proto/frequenz/api/microgrid/v1/microgrid.proto
tiyash-basu-frequenz: Done

proto/frequenz/api/microgrid/v1/microgrid.proto
tiyash-basu-frequenz: Good catch! Fixed

proto/frequenz/api/microgrid/v1/microgrid.proto
tiyash-basu-frequenz: Done

hannah-stevenson-frequenz (approved):

PR #230, opened 2024-03-27
ci: Use a hash for the labeler version instead of a tag

llucax:
For security reasons, since we are allowing access to secrets, we should use a hash instead of a tag.


llucax:
Enabling auto-merge.

Marenz (approved):

tiyash-basu-frequenz (approved):

PR #231, opened 2024-03-28
Allow clients to specify validity durations for bounds

### LINKED ISSUE #181: Allow clients to specify durations for provided bounds
### What's needed?

Currently, the bounds provided by clients get invalidated after 5 seconds by default. The API could be extended to allow clients specify their own durations for this.

### Proposed solution

_No response_

### Use cases

At the moment, the default duration for bounds to stay valid is 5 seconds. Which means users need to keep sending bounds, otherwise they will lapse. Assuming a user needs to apply the same pair of bounds for the next 15 minutes, then they would need the same pair of bounds at least 180 times for it to be valid over the whole 15 minute window. Having the ability to provide custom validity durations will help reducing this overhead.

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
This commit adds a field named `validity_duration` to the `AddComponentBoundsRequest` message, which can be used by clients to specify a custom default duration to bounds while adding them. If not specified, a default validity duration of 5 seconds is applied.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -17,11 +17,13 @@
 - The enum `ComponentBoundsTargetMetric` has been removed in favour of the
   `Metric` enum from `frequenz-api-common`.
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
-
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+- The `AddComponentBoundsRequest` message has a field `validity_duration` which
+  allows the user to specify the duration for which the bounds are valid. The
+  bounds will be automatically removed after the specified duration. The client
+  can select between 5 seconds, 1 minute, 5 minutes, and 15 minutes. If set to
+  `UNSPECIFIED`, the bounds will be valid for a default duration of 5 seconds.
 
 ## Bug Fixes
 
--- a/proto/frequenz/api/microgrid/v1/microgrid.proto
+++ b/proto/frequenz/api/microgrid/v1/microgrid.proto
@@ -131,7 +131,11 @@ service Microgrid {
   }
 
   // Adds inclusion bounds for a given metric of a given component. Returns the
-  // time at which the bounds will expire (as a timestamp (UTC))
+  // time at which the bounds will expire (as a timestamp (UTC)).
+  //
+  // The request parameters allows users to select a duration until
+  // which the bounds will stay in effect. If no duration is provided, then the
+  // bounds will be removed after a default duration of 5 seconds.
   //
   // Inclusion bounds give the range that the system will try to keep the
   // metric within. If the metric goes outside of these bounds, the system will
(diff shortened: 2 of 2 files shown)


tiyash-basu-frequenz:
The context being that we do not want clients to set bounds for arbitrarily long durations:
An alternative approach to the one in this PR would be to allow the client to ask for an arbitrary time, but limit it internally to 15 minutes (and of course, document it). I did not go for that approach because that approach requires users to go through the docs, whereas this approach enforces such limits directly in the API.

Would be interested in hearing your opinions.

llucax:
>  An alternative approach to the one in this PR would be to allow the client to ask for an arbitrary time, but limit it internally to 15 minutes (and of course, document it). I did not go for that approach because that approach requires users to go through the docs, whereas this approach enforces such limits directly in the API.

I agree it is better to fail if the requested time can't be fulfilled, otherwise it could lead to unexpected behaviour. But any reason not to use an arbitrary number, document a maximum and minimum allowed value and fail the request if the value is out of the valid range? For me it looks a bit weird to have these times as an enum.

tiyash-basu-frequenz:
>But any reason not to use an arbitrary number, document a maximum and minimum allowed value and fail the request if the value is out of the valid range? For me it looks a bit weird to have these times as an enum.

If it is an enum, users will get a compilation error if they choose an invalid timestamp. If itis  a number, then they will get a runtime error, which could mean more mistakes and more error-handling (so more code). Hence my first choice was an enum.

But TBH, I do not have a strong opinion either way. I figured I could just create the PR and discuss it here.

llucax:
I'm not completely against the enum either, which I guess it is a bad thing, as it is harder to decide :P

I just find it a bit weird, but I never used it either, so maybe I'm just concerned that it might be a bit inflexible but it is an unjustified concern and it might be more than enough flexibility. So probably it is better to get someone that really use bounds to talk. Maybe @shsms @daniel-zullo-frequenz @Marenz ?

shsms:
The SDK won't be sending bounds once we upgrade to this version, just setting power.  For power, the interval is not customizable right?  If it is customizable, we'd have to get an option from the actors, so that the SDK doesn't have to make generalized decisions.

Also, I'm curious about the motivation for asking the users to decide.  Why/when would I as a user choose 15 minutes over the default 5 seconds, and repeat commands every 3 seconds?  Are there any specific usecases/requirements that's driving this?

tiyash-basu-frequenz:
>The SDK won't be sending bounds once we upgrade to this version, just setting power. For power, the interval is not customizable right? If it is customizable, we'd have to get an option from the actors, so that the SDK doesn't have to make generalized decisions.

The idea is to make it customizable.

>Also, I'm curious about the motivation for asking the users to decide. Why/when would I as a user choose 15 minutes over the default 5 seconds, and repeat commands every 3 seconds? Are there any specific usecases/requirements that's driving this?

Good point, I forgot to link the issue. It is linked now, and has the use-case mentioned there.

tiyash-basu-frequenz:
ping

thomas-nicolai-frequenz (approved):
LGTM

stefan-brus-frequenz (approved):
LGTM

shsms (approved):
LGTM

thea-leake (approved):
LGMT!

hannah-stevenson-frequenz (approved):

PR #235, opened 2024-04-09
Fix python dependencies for frequenz-api-common v0.6.0

tiyash-basu-frequenz:


Marenz (approved):

thea-leake (approved):
LGTM!

pyproject.toml
```diff
@@ -26,8 +26,8 @@ classifiers = [
 ]
 requires-python = ">= 3.11, < 4"
 dependencies = [
-  "frequenz-api-common >= 0.5.0, < 0.6",
-  "googleapis-common-protos >= 1.56.2, < 2",
+  "frequenz-api-common >= 0.6.0, < 0.7",
+  "googleapis-common-protos >= 1.56.4, < 2",
```
llucax, line 30: I recommend widening dependencies for libraries instead of bumping whenever possible. For `frequenz-api-common` I guess we need to bump because they are incompatible, but I guess we can still support `googleapis-common-protos` 1.56.2, right?

tiyash-basu-frequenz: I was getting the follwing error with the version you suggest:
```
ERROR: Cannot install frequenz-api-common==0.6.0 and googleapis-common-protos==1.56.2 because these package versions have conflicting dependencies.

The conflict is caused by:
    frequenz-api-common 0.6.0 depends on protobuf<5 and >=4.25.3
    googleapis-common-protos 1.56.2 depends on protobuf<4.0.0dev and >=3.15.0
```

tiyash-basu-frequenz: Here is the commit where `protobuf` was updated in frequenz-api-common: https://github.com/frequenz-floss/frequenz-api-common/commit/b4c5e62abb3533b6c1e790ec8bb10bd2497494ab

llucax: Mmm, weird, `frequenz-api-common` doesn't depend on `googleapis-common-protos` anymore, but we added the dependency to `protobuf >= 4.25.3` to do that, I guess `googleapis-common-protos` depends 1.56.2 depends on an older `protobuf` and this is where the conflict is coming from. I guess I should have used an older `protobuf` version.

We could fix `api-common` but I guess it is too much work and we can do the bump here for now. Dependencies are very tricky :grimacing:

shsms: > Dependencies are very tricky

I agree, we end up depending on them, compromising our independence. it is nonsense.

llucax (approved):
I think it is not worth updating `api-common` because of this, so all good to merge from my side.

PR #237, opened 2024-04-12
Add CI step to build using `protoc` with fatal warnings

### LINKED ISSUE #228: Report unused imports in CI
### What's needed?

We have seen at least once that our protobuf files had unused imports. This did not cause any issue with Rust or Python toolcahins, but caused issues with a Go toolchain (used for building the gRPC gateway). This needed to be [fixed](https://github.com/frequenz-floss/frequenz-api-microgrid/pull/225).

We need the CI to catch such unused imports, so that
a. our toolchains do not break
b. our proto definitions stay clean

The CI should also fail on finding unused imports.

### Proposed solution

The first solution that comes to mind is to run protoc to generate Go code, since that definitely broke. Howeverm there could be other solutions, including the linter. We need to check.

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
The steps in the CI file from the step `nox` onwards miss a few cases, like reporting unused imports. An additional step with `protoc` takes care of this. This step generates the protoc files directly using `protoc`, and treats warnings as fatal.

Closes #228

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,6 +46,32 @@ jobs:
           protolint_version: "0.45.0"
           reporter: github-check
 
+  protoc:
+    name: Test with protoc
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Fetch sources
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: Source protoc
+        run: |
+          curl -LO https://github.com/protocolbuffers/protobuf/releases/download/v25.1/protoc-25.1-linux-x86_64.zip
+          mkdir protoc
+          unzip protoc-25.1-linux-x86_64.zip -d protoc
+          sudo cp -r protoc/bin/* /usr/bin/
+          sudo cp -r protoc/include/* /usr/include/
+          rm -r protoc-25.1-linux-x86_64.zip protoc
+      - name: Run sanity tests
+        run: |
+          protoc \
+          --fatal_warnings \
+          -I proto \
+          -I submodules/frequenz-api-common/proto \
+          -I submodules/api-common-protos \
+          --python_out=. \
+          frequenz/api/microgrid/v1/microgrid.proto
+
   nox:
     name: Test with nox
     strategy:
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -38,4 +38,6 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+- The CI was unable to catch unused imports in the proto file before. This has
+  been fixed by adding a new step to the CI to build using `protoc` with the
+  `--fatal-warnings` flag.


matthias-wende-frequenz (approved):
Looks good but I wonder if that checks should be moved to nox and repo-config. @llucax Opinions?

.github/workflows/ci.yaml
```diff
@@ -46,6 +46,32 @@ jobs:
...
+      - name: Fetch sources
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: Source protoc
+        run: |
+          curl -LO https://github.com/protocolbuffers/protobuf/releases/download/v25.1/protoc-25.1-linux-x86_64.zip
+          mkdir protoc
+          unzip protoc-25.1-linux-x86_64.zip -d protoc
+          sudo cp -r protoc/bin/* /usr/bin/
```
hannah-stevenson-frequenz, line 62: A minor point, given this is all happening in a throw away container, but cping into /usr seems messy. Maybe this stuff should go  in it's own directory and paths updated by tools looking for it.

Given we will only ever have one version and the container will be gone a minute later though just ignore me - its not worth it.

hannah-stevenson-frequenz (approved):

thea-leake (approved):
LGMT!

tiyash-basu-frequenz:
>Looks good but I wonder if that checks should be moved to nox and repo-config. @llucax Opinions?

Right, this should be added to repo-config at some point. Also, the action for setting up protoc has to be extracted into a separate action.

.github/workflows/ci.yaml
tiyash-basu-frequenz: This will be gone away after this step. The next step gets a different ubuntu environment. Not worth changing it, I'd say.

llucax: I wonder why not just using the ubuntu package `protobuf-compiler` (using a newer ubuntu base image probably), or just using an official docker protobuf docker image (https://hub.docker.com/u/grpc).

llucax (commented):
Just a minor comment, but I'm also fine to keep it as it is.

llucax:
> Looks good but I wonder if that checks should be moved to nox and repo-config. @llucax Opinions?

yes, but no, but yes

repo-config is needing some love, we are not being able to catch-up with the pace of new repo types (like api clients), and I want to split it in smaller repos (one for generating protobuf, one with the tools for mkdocs, one with just the cookicutter templates, etc.). We want to remove the python code generation too soon :tm: from "api" repos, so there is a lot of work to do on that side too. So I think we can wait for all that restructuring before pushing this to repo-config. Eventually I think we'll need to do a sprint to *re-repo-configy* all repositories.

llucax:
> Right, this should be added to repo-config at some point. Also, the action for setting up protoc has to be extracted into a separate action.

Yeah, that too, I started the process of splitting the big `ci.yaml` workflow in repo-config in smaller GitHub actions! But didn't have the time to update the cookiecutter templates to use those yet (which are also probably not done).

llucax:
So we don't forget:

* https://github.com/frequenz-floss/frequenz-repo-config-python/issues/238

.github/workflows/ci.yaml
hannah-stevenson-frequenz: The one that comes with ubuntu 20 is very old (3.something).  Though as you say, we could use a different image.

tiyash-basu-frequenz: > I wonder why not just using the ubuntu package protobuf-compiler

Then we get stuck with whatever Ubuntu provides, which generally gets outdated very quickly.

>  or just using an official docker protobuf docker image

Downloading the binary from gh did not feel like a hassle TBH. But containers could certainly be worth a try.

PR #254, opened 2024-09-10
Pin grpc related dependencies

llucax:
We can't use wide dependencies because otherwise when building the wheel, the latest version is used, but the generate code should use the minimum supported version for the generation, not the latest one. See: protobuf.dev/support/cross-version-runtime-guarantee

We also need to update the `frequenz-api-common` dependency to get the fix for the `grpcio` version too.


llucax:
Will force merge as things are broken as they are. We can fix anything that needs fixing in a future PR.

PR #261, opened 2024-09-18
Bump `protobuf` and `grpcio` dependencies

llucax:
The `protobuf` version 4.25.x will go out of support soon(ish), on 31 Mar 2025. Since we don't need to keep compatibility to any legacy code, we can bump it in advance as a preparation.

The `googleapis-common-protos` version was also bumped to 1.65.0, as the previous version was too old to work with the new `protobuf` version.

https://protobuf.dev/support/version-support/#python

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,7 +6,7 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+- The minimum supported version of `protobuf` was bumped to 5.28.1 and to 1.66.1 for `grpcio`. Make sure to update your dependencies accordingly.
 
 ## New Features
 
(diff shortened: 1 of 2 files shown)


llucax:
This is technically a breaking change, so we might want to wait to merge it if there are any planned fix releases for v0.17.x.

daniel-zullo-frequenz (approved):

tiyash-basu-frequenz (approved):

llucax:
@tiyash-basu-frequenz does the approval means it is ok to merge now?

tiyash-basu-frequenz:
Yep, it's ok to merge it, since it does not touch proto files.

llucax:
But Python-wise is technically a breaking change, so it shouldn't be released as v0.17.x. This is where having binding and spec files start to get really messy :grimacing: 

I guess since we are in pre-1.0 and it is a minor breaking change (you just need to bump deps) it might be OK to make an exception and release in v0.17.x if needed too.

PR #295, opened 2025-04-15
Add Test with protoc to the Test PR workflows

llucax:
This job is not yet provided by repo-config, so we need to add it manually.


llucax:
Will force-merge as it is a copy&paste from a job we are already using.

PR #312, opened 2025-06-10
Bump protobuf-related dependencies

llucax:
Also ignore warnings about protobuf gencode version, as it is guaranteed by protobuf that 2 consecutive major versions are compatible. We are aware that after that we need to update the depencencies, and we'll get an error if we forget anyways.


llucax:
Routine update applied to many repos, so also force-merging after tests are done.

PR #323, opened 2025-07-01
dependabot: Group grpc/protobuf updates together

llucax:
We need to do this because we need to manually update these to ensure the cross-version guarantees between grpcio and protobuf are respected.


llucax:
Will implement this in repo-config soon.

tiyash-basu-frequenz (approved):

PR #331, opened 2025-07-07
Add `ramp_rate` to SetComponentPowerActive/ReactiveRequest

### LINKED ISSUE #287: Allow clients to specify ramp rates in set-power commands
### What's needed?

Different clients may have a need for different ramp rates when setting active or reactive power. Currently, ramp rates are applied at the discretion of the service, which does not allow the user to customise it. We want to fix this situation.

### Proposed solution

The most trivial solution would be to take in an optional ramp rate in the set-power RPCs. When provided, this should be a positive integer, in watts/second. When not provided, it should default to 10% of the maximum active or reactive power, based upon the invoked RPC. The ramp should be applied linearly by the service.

### Use cases

The main use-case of this would be in cases where the client needs to be able to tune ramp rates, in order to maintain stable operations, e.g., ramping power slowly for grid-compliance, or ramping power quickly for [S|P]RL.

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
This commit adds a `ramp_rate` field to the
`SetComponentPowerActiveRequest` and `SetComponentPowerReactiveRequest` messages. This field allows users to specify the desired rate of change for the power output of electrical components in a microgrid. This is useful in scenarios where the rate of change of power is critical tio the use-case of the client.

The `ramp_rate` field is an optional unsigned integer that represents the magnitude of the ramp in watts per second (for active power) or volt-amperes per second (for reactive power). The service will apply the ramp in the correct direction (up or down) to reach the target power. The actual rate of change will be capped by the component's configured maximum ramp rate, ensuring that the system operates within the component's limits.

If the `request_lifetime` ends before the ramp is complete, the component will be set to its standby state. If a new command arrives mid-ramp, the current ramp will be interrupted, and a new one will begin according to the new command's parameters. If the `ramp_rate` is not provided, the component will use its configured maximum ramp rate.

closes #287

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -23,6 +23,7 @@
   - `ReceiveSensorDataStreamResponse` RPC method has been renamed to `ReceiveSensorTelemetryStreamResponse` to match the naming conventions used in `frequenz-api-common`.
 - The RPC `GetMicrogridMetadata` has been renamed to `GetMicrogrid` to better align with the naming convention in `frequenz-api-common`.
 - The `GetMicrogridMetadataResponse` message has been renamed to `GetMicrogridResponse` to better align with the naming convention in `frequenz-api-common`.
+- Add `ramp_rate` to `SetElectricalComponentPowerActiveRequest` and `SetElectricalComponentPowerReactiveRequest` messages to allow setting the desired ramp rate for active and reactive power changes.
 
 ## New Features
 
--- a/proto/frequenz/api/microgrid/v1/microgrid.proto
+++ b/proto/frequenz/api/microgrid/v1/microgrid.proto
@@ -450,7 +450,6 @@ message ReceiveSensorTelemetryStreamResponse {
   // The sensor data.
   frequenz.api.common.v1alpha7.microgrid.sensors.SensorTelemetry telemetry = 1;
 }
-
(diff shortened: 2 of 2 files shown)


thomas-nicolai-frequenz:
@tiyash-basu-frequenz to me this feels like desired or max ramp rate. How does the user know what the max ramp rate is the component does support? The user could try to set something thats not feasible.

tiyash-basu-frequenz:
> to me this feels like desired or max ramp rate.

It is exactly that.

> How does the user know what the max ramp rate is the component does support?

This is not something that is addressed in this PR yet. The way to address this would be to add ramp rates as metrics to `frequenz-api-common`.

> The user could try to set something thats not feasible.

The current approach described in this PR is that we allow the command, but cap the ramp rate to the maximum.

thomas-nicolai-frequenz:
> The current approach described in this PR is that we allow the command, but cap the ramp rate to the maximum.

Does the API inform the user about the clipping / applied ramp rate? Btw what use cases require this? Is this urgent? Otherwise the question is if we postpone that after v1?

tiyash-basu-frequenz:
@thomas-nicolai-frequenz and the rest of us discussed this issue elsewhere. We will explore other solutions to provide this feature. So for now, this PR can be closed.

PR #332, opened 2025-07-07
Stream responses for `SetElectricalComponentPowerActive/Reactive` RPCs 

### LINKED ISSUE #251: Enhancing Hardware Command Responses with Detailed Asynchronous Status Updates
### What's needed?

We need a more informative way to get responses from hardware components after issuing commands to them.

Currently, the response is merely an acknowledgement and does not reflect the actual state or events at the component. Ideally, responses should include the statuses of all activities our stack performs on the component.

For example, the backend might increase power to 100kW in a 3-step ramp of 10kW, 50kW, and 100kW. Currently, the response would only acknowledge the 100kW command. If the 50kW command fails when sent to the inverter (e.g., due to sudden bounds change or communication errors), the client remains unaware and has to rely solely on the power value in the component data stream. This is suboptimal.

This may involve exposing more low-level information to the clients, so we need to make this optional. Clients who do not want this information should not be required to receive it.

### Proposed solution

There are two ways of addressing this:

### Proposal 1

Update the set-power command RPCs to return a stream of responses.

### Proposal 2

Have a separate RPC that streams command response statuses. In this case, each command should be provided a unique ID from the service, and the streamed response status should contain this ID. (although one client should not get response for commands from another client).

Both proposals are technically feasible from the service's perspective. However, we need feedback from current users to determine their preferences and reasons.

### Use cases

With the proposed changes, the client would receive a stream of real-time status updates for each command.

For example, after issuing the command to increase power, the client would receive updates indicating the progress and success or failure of each step in the process. If the command to write to the Modbus register succeeds but the component fails to respond within the expected timeframe, the client would be immediately informed of the issue, allowing for timely corrective actions such as retrying the command, diagnosing the communication problem, or adjusting their operational strategy based on the current status of the component.

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
The response messages for the `SetElectricalComponentPowerActive` and `SetElectricalComponentPowerReactive` RPCs have been changed to return a stream of responses instead of a single response. This allows the server to provide ongoing status updates for the request, which is useful when the electrical component cannot immediately set its output power to the requested value. The initial response will indicate whether the request was accepted or rejected, and a subsequent response will indicate whether the request was successful, failed, or overridden by another request. If the request was accepted and successful, the response will also contain the timestamp until which the given power command will stay in effect, after which the electrical component will be returned to its default state.

### DIFF
--- a/proto/frequenz/api/microgrid/v1/microgrid.proto
+++ b/proto/frequenz/api/microgrid/v1/microgrid.proto
@@ -169,11 +169,19 @@ service Microgrid {
   // The power output can be -ve or +ve, depending on whether the electrical
   // component is supposed to be discharging or charging, respectively.
   //
-  // The return value is the timestamp until which the given power command will
-  // stay in effect. After this timestamp, the electrical component's active
-  // power will be set to 0, if the API receives no further command to change
-  // it before then. By default, this timestamp will be set to the current time
-  // plus 60 seconds.
+  // The server provides a stream of responses in return. Each response will
+  // contain the current status of the request.
+  // **Initial response:** The initial response will be sent immediately after
+  // the request is received, and will indicate whether the request was accepted
+  // or rejected. If the request was accepted, this response will also contain
+  // the timestamp until which the given power command will stay in effect,
+  // after which the electrical component will be returned to its default state.
+  // **Subsequent response:** After the initial response, the server will send
+  // another response in the stream, which will be the final response. It will
+  // indicate whether the request was successful, failed, or overridden by
+  // another request. If the request was successful, it will also contain the
+  // timestamp until which the given power command will stay in effect, after
+  // which the electrical component will be returned to its default state.
   //
   // Note that the target electrical component may have a resolution of more
   // than 1 W. E.g., an inverter may have a resolution of 88 W.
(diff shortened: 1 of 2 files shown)


thomas-nicolai-frequenz (dismissed):
LGTM

thomas-nicolai-frequenz (approved):
LGTM

proto/frequenz/api/microgrid/v1/microgrid.proto
```diff
@@ -497,13 +513,78 @@ message SetElectricalComponentPowerActiveRequest {
...
-  // be set to the current time plus 60 seconds.
-  google.protobuf.Timestamp valid_until = 1;
+  // After this timestamp, the electrical component power will be set to its
+  // standby state, if the API receives no further power commands.
+  // By default, this timestamp will be set to the current time plus 60 seconds.
+  //
+  // When the request status is `REJECTED`, `FAILED`, or `OVERRIDDEN`, the
+  // `valid_until` timestamp will not be set, and the electrical component
+  // will not change its power output.
+  optional google.protobuf.Timestamp valid_until = 1;
```
llucax, line 583: Nitpick: this `optional` is redundant as `Timestamp` is a `message` and all `messages` are optional. But I guess documentation wise it can be nice to make it more explicit that this field can be absent, while other `messages` types technically can also be absent and the protobuf message will be (de)serialized without errors, but it still should be an error at the user-level.

Are we doing this consistently? Because I found myself many time asking myself if a field with `message` type is "required" or not (i.e. if I can just discard the message as invalid if the field is missing) or if I should just read the default values.

tiyash-basu-frequenz: true, it does not need the `optional` keyword. I will remove it.

thomas-nicolai-frequenz: Yes, its only required for simple types like string or boolan.

tiyash-basu-frequenz: https://github.com/frequenz-floss/frequenz-api-microgrid/pull/333

PR #335, opened 2025-07-08
Rename fields in `ListElectricalComponentConnectionsRequest`

### LINKED ISSUE #262: Fix terminology in the `ListConnections` RPC
### What's needed?

The `ListConnections` RPC has a request message that talks about `start` and `end`, which is a bit confusing, as it sounds like it might be referring to ranges of IDs, which is very weird.

https://github.com/frequenz-floss/frequenz-api-microgrid/blob/8f4e4f897d0082140c8eb8da9dd70c8efdb1aa92/proto/frequenz/api/microgrid/v1/microgrid.proto#L360-L370

The [ `ComponentConnection` message in `api-common`](https://github.com/frequenz-floss/frequenz-api-common/blob/ca56060fbf55975299233aa6af15b6dfa6e77bc6/proto/frequenz/api/common/v1/microgrid/components/components.proto#L192-L200) was renamed to use `source` and `destination` instead of `start` and `end`, so the microgrid API should probably adopt those terms too.

### Proposed solution

```protobuf
// Request parameters for the RPC `ListConnections`.
// Contains filtering parameters for listing connections.
message ListConnectionsRequest {
  // Only return connections that originate from the specified component ID(s):
  // if empty, connections with any `source_component_id` will be returned
  repeated uint64 source_component_ids = 1;

  // Only return connections that terminate at the specified component ID(s):
  // if empty, connections with any `destination_component_id` will be returned
  repeated uint64 destination_component_ids = 2;
}
```

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
This commit renames the fields in the
`ListElectricalComponentConnectionsRequest` message:
- `starts` is renamed to `source_component_ids`
- `ends` is renamed to `destination_component_ids`

This change improves clarity by explicitly indicating that the fields represent the source and destination electrical component IDs in the connections request.

This change also aligns it with the naming convention of electrical component connections in the `frequenz-api-common` repository.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -21,6 +21,8 @@
   - `ReceiveSensorDataStream` RPC method has been renamed to `ReceiveSensorTelemetryStream` to match the naming conventions used in `frequenz-api-common`.
   - `ReceiveSensorDataStreamRequest` RPC method has been renamed to `ReceiveSensorTelemetryStreamRequest` to match the naming conventions used in `frequenz-api-common`.
   - `ReceiveSensorDataStreamResponse` RPC method has been renamed to `ReceiveSensorTelemetryStreamResponse` to match the naming conventions used in `frequenz-api-common`.
+  - `ListComponentConnectionRequests.starts` has been renamed to `ListElectricalComponentConnectionRequests.source_component_ids` to match the naming conventions used in `frequenz-api-common`.
+  - `ListComponentConnectionRequests.ends` has been renamed to `ListElectricalComponentConnectionRequests.destination_component_ids` to match the naming conventions used in `frequenz-api-common`.
 - The RPC `GetMicrogridMetadata` has been renamed to `GetMicrogrid` to better align with the naming convention in `frequenz-api-common`.
 - The `GetMicrogridMetadataResponse` message has been renamed to `GetMicrogridResponse` to better align with the naming convention in `frequenz-api-common`.
 - The response messages for the `SetElectricalComponentPowerActive` and
--- a/proto/frequenz/api/microgrid/v1/microgrid.proto
+++ b/proto/frequenz/api/microgrid/v1/microgrid.proto
@@ -383,13 +383,15 @@ message ListSensorsResponse {
 message ListElectricalComponentConnectionsRequest {
   // Only return connections that start from the specified electrical
   // component ID(s).
-  // If empty, connections with any `start` will be returned.
-  repeated uint64 starts = 1;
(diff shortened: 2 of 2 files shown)


llucax (approved):

llucax:
Actually, if we are going full names shouldn't this be `xxx_electrical_component_id`? From the context the `electrical_` part might be redundant but I think we were adding it in other contexts, so maybe for consistency we should add it everything.

Removing from the queue until we decide.

llucax:
For example `ListElectricalComponentsRequest` uses `electrical_component_ids`.

tiyash-basu-frequenz:
In the common API, we use `source|destination_component_id` to denote [these](https://github.com/frequenz-floss/frequenz-api-common/blob/7953a913f3cf4954fb484e02007d82e71c045925/proto/frequenz/api/common/v1alpha7/microgrid/electrical_components/electrical_components.proto#L656-L668).

So while the changes here are not verbose, it is in line with the definitions in the common API.

I do not have any preference here. If you have a preference, I will be happy to change it.

llucax:
As you might know by now, I like consistency. I hate long names, but I like consistency more than I hate long names. I think it is more consistent and clear if we refer to electrical component IDs in the same way across all the APIs, so I would be in favor of renaming all references to electrical component IDs as `electrical_component_id` both in this and the common API :grimacing:

llucax:
I'm just seeing we also [have a plain `component_id ` in the telemetry message](https://github.com/frequenz-floss/frequenz-api-common/blob/7953a913f3cf4954fb484e02007d82e71c045925/proto/frequenz/api/common/v1alpha7/microgrid/electrical_components/electrical_components.proto#L793) (and a [wrong example](https://github.com/frequenz-floss/frequenz-api-common/blob/7953a913f3cf4954fb484e02007d82e71c045925/proto/frequenz/api/common/v1alpha7/microgrid/electrical_components/electrical_components.proto#L756)). I would *fix* those too.

llucax:
So, wrapping up a bit. Since this microgrid API version uses v1alpha7, I'm good with merging this as is, as it is consistent with the common API, so for me both options are equally good/bad as we'll have some inconsistency in this version, but it would be nice to have `electrical_component_id` everywhere for v1alpha8/v0.19.

tiyash-basu-frequenz:
As discussed, we will updating the `ElectricalComponentConnection` message in the upcoming fr4equenz-api-common release: https://github.com/frequenz-floss/frequenz-api-common/issues/395.

Will merge this PR for now.

PR #336, opened 2025-07-08
Unify power setting RPCs into new RPC `SetElectricalComponentPower`

tiyash-basu-frequenz:
This commit adds a new RPC `SetElectricalComponentPower` to the Microgrid API. This RPC allows clients to set either the active or reactive power of an electrical component, depending on the value of the `power_type` field in the request message. The request message includes the target power in watts (W) for active power and in volt-amperes reactive (VAR) for reactive power, with appropriate sign conventions.

The server responds with a stream of responses, providing ongoing status updates for the request. The initial response indicates whether the request was accepted or rejected, and if accepted, it includes a timestamp until which the power command will remain in effect. The subsequent response indicates the final status of the request, including whether it was successful, failed, or overridden by another request, and if successful, it also includes the timestamp until which the power command will remain in effect.

Since the `SetElectricalComponentPower` RPC has been added, the `SetElectricalComponentPowerActive` and `SetElectricalComponentPowerReactive` RPCs are no longer needed. They have been removed to simplify the API and avoid redundancy.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -28,6 +28,7 @@
 - The response messages for the `SetElectricalComponentPowerActive` and
 `SetElectricalComponentPowerReactive` RPCs have been changed to return a stream of responses instead of a single response. This allows the server to provide ongoing status updates for the request, which is useful when the electrical component cannot immediately set its output power to the requested value.
 - The RPC `AddComponentBounds` has been renamed to `AugmentElectricalComponentBounds`.
+- A new RPC `SetElectricalComponentPower` has been added. It allows setting either the active or reactive power of an electrical component, depending on the value of the `power_type` field in the request message.
 
 ## New Features
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -28,7 +28,7 @@
 - The response messages for the `SetElectricalComponentPowerActive` and
 `SetElectricalComponentPowerReactive` RPCs have been changed to return a stream of responses instead of a single response. This allows the server to provide ongoing status updates for the request, which is useful when the electrical component cannot immediately set its output power to the requested value.
 - The RPC `AddComponentBounds` has been renamed to `AugmentElectricalComponentBounds`.
-- A new RPC `SetElectricalComponentPower` has been added. It allows setting either the active or reactive power of an electrical component, depending on the value of the `power_type` field in the request message.
+- `SetElectricalComponentPowerActive` and `SetElectricalComponentPowerReactive` RPCs ave been replaced by a new RPC `SetElectricalComponentPower`. The new RPC allows setting either the active or reactive power of an electrical component, depending on the value of the `power_type` field in the request message.
 
 ## New Features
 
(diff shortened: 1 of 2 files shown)


proto/frequenz/api/microgrid/v1/microgrid.proto
```diff
@@ -165,6 +165,61 @@ service Microgrid {
...
+  // which the electrical component will be returned to its default state.
+  //
+  // This command is supported for the following electrical component
+  // categories, provided the specific model also supports it:
+  // - `ELECTRICAL_COMPONENT_CATEGORY_CAPACITOR_BANK`
+  // - `ELECTRICAL_COMPONENT_CATEGORY_CHP`
+  // - `ELECTRICAL_COMPONENT_CATEGORY_ELECTROLYZER`
+  // - `ELECTRICAL_COMPONENT_CATEGORY_EV_CHARGER`
+  // - `ELECTRICAL_COMPONENT_CATEGORY_INVERTER`
+  // - `ELECTRICAL_COMPONENT_CATEGORY_WIND_TURBINE`
```
llucax, line 215: I don't remember if this component type is already gone is it will be gone in a future PR.

RELEASE_NOTES.md
```diff
@@ -26,6 +26,7 @@
 - The response messages for the `SetElectricalComponentPowerActive` and
 `SetElectricalComponentPowerReactive` RPCs have been changed to return a stream of responses instead of a single response. This allows the server to provide ongoing status updates for the request, which is useful when the electrical component cannot immediately set its output power to the requested value.
 - The RPC `AddComponentBounds` has been renamed to `AugmentElectricalComponentBounds`.
+- A new RPC `SetElectricalComponentPower` has been added. It allows setting either the active or reactive power of an electrical component, depending on the value of the `power_type` field in the request message. This replaces the previous `SetElectricalComponentPowerActive` and `SetElectricalComponentPowerReactive` RPCs.
```
llucax: As upgrading instructions It might be more clear to reverse this, and say something like "`SetElectricalComponentPowerActive` and `SetElectricalComponentPowerReactive` RPCs were replaced by `SetElectricalComponentPower` ..."

llucax (dismissed):
I find the PR title is a bit misleading, I would change it to "Unify power setting RPCs" or something like that, as it is not only adding a RPC, is removing a couple too.

But all comments are minor, so approving in case you want to merge as is.

proto/frequenz/api/microgrid/v1/microgrid.proto
tiyash-basu-frequenz: IMO it is better to keep it as a component category and remove it from the inverter type enum: https://github.com/frequenz-floss/frequenz-api-common/issues/393

So I'd keep this as is for now. If things change i the common API v0.8.0 release, we can change this one too.

RELEASE_NOTES.md
tiyash-basu-frequenz: Done.

tiyash-basu-frequenz:
>I find the PR title is a bit misleading, I would change it to "Unify power setting RPCs" or something like that, as it is not only adding a RPC, is removing a couple too.

Done.

llucax (approved):

PR #382, opened 2026-03-03
Add power-source selection for hybrid inverter controls

### LINKED ISSUE #258: Allow Specifying DC Energy Source for Hybrid Inverters
### What's needed?

So far, the RPC [`SetComponentPowerActive`](https://github.com/frequenz-floss/frequenz-api-microgrid/blob/6291e8e9d1c0769ec7b93cb8300dfb3c3f096ec0/proto/frequenz/api/microgrid/v1/microgrid.proto#L190-L195) only allows setting the total active power of a component without differentiating between the sources (PV arrays and batteries) when dealing with hybrid inverters.

This limitation primarily affects scenarios where more granular control is required, especially during discharging operations, as clients may need to specify whether the energy is drawn from the battery or the PV arrays.

We need a way to specify the DC energy source when issuing discharge commands.

We do not need to specify the DC component for
- charge commands, because it will anyway always target batteries, as there is no charge functionality for PV arrays.
- reactive-power commands, because inverters manage reactive power generation or absorption internally.

#### Proposal 1

This approach is about asking the user specifics for hybrid-inverter-discharges, and ignoring this input for other cases.

```proto
message SetComponentPowerActiveRequest {
  ...

  // This is the field being added.
  // For PV and battery inverters, this field will be ignored
  // For hybrid inverters, for charge commands, this field will be ignored.
  // For hybrid inverters, for discharge commands, this field is mandatory.
  ComponentPowerSource source = 4;
}

enum ComponentPowerSource {
  // Default
  COMPONENT_POWER_SOURCE_UNSPECIFIED = 0;

  // When the PV arrrays should be used as the energy source.
  COMPONENT_POWER_SOURCE_PV = 1;

  // When the batteries should be used as the energy source.
  COMPONENT_POWER_SOURCE_BATTERY = 2;
}
```

#### Proposal 2

This approach is about using defaults as much as possible.

```proto
message SetComponentPowerActiveRequest {
  ...

  // This is the field being added.
  // For PV inverters, specifying battery will result in an error.
  // For battery inverters, specifying PV will result in an error.
  // For hybrid inverters,
  //   - specifying PV for charging will result in an error.
  //   - specifying nothing for discharging will result in preferring the PV, and any excess goes to batteries.
  //   - specifying PV or batteries will result in the input being chosen as the preferred source.
  ComponentPowerPreferredSource source = 4;
}

enum ComponentPowerPreferredSource {
  // Default
  COMPONENT_POWER_PREFERRED_SOURCE_UNSPECIFIED = 0;

  // When the PV arrays should be preferred as the energy source.
  COMPONENT_POWER_PREFERRED_SOURCE_PV = 1;

  // When the batteries should be preferred as the energy source.
  COMPONENT_POWER_PREFERRED_SOURCE_BATTERY = 2;
}
```

#### Proposal 3

This is a combination of the above two:

```proto
message SetComponentPowerActiveRequest {
  ...

  // This is the field being added.
  // For PV inverters, specifying battery will result in an error.
  // For battery inverters, specifying PV will result in an error.
  // For hybrid inverters,
  // ...
  ComponentPowerSource source = 4;
}

enum ComponentPowerSource {
  // Default. Results in PV being preferred.
  COMPONENT_POWER_SOURCE_UNSPECIFIED = 0;

  // When only the PV arrays should be used as the energy source.
  // Commands exceeding the limits of the PV arrays will result in an error.
  COMPONENT_POWER_SOURCE_PV_ONLY = 1;

  // When the PV arrays should be used as the preferred energy source,
  // and any leftover power is pulled from batteries.
  COMPONENT_POWER_SOURCE_PV_PREFERRED = 2;

  // When only the batteries should be used as the energy source.
  // Commands exceeding the limits of the batteries will result in an error.
  COMPONENT_POWER_SOURCE_BATTERY_ONLY = 3;

  // When the batteries should be used as the preferred energy source,
  // and any leftover power is pulled from PV, as much as possible.
  COMPONENT_POWER_SOURCE_BATTERY_PREFERRED = 2;
}
```

One implication of this approach is that the underlying controller should be responsible for maintaining a steady AC output if the PV power fluctuates.

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
This commit adds the ability to specify a preferred power source when issuing discharge commands to hybrid inverters, which typically have both PV and battery inputs added to them, and can use either of those inputs as the power source.

Both the `v1alpha18` and `v1alpha19` APIs are updated in this PR. Since `v0.19.0` is not yet released, I have renumbered the fields in the `SetElectricalComponentPowerRequest` message of the `v1alpha19` package.

closes #258

### DIFF
--- a/proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
+++ b/proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
@@ -497,6 +497,44 @@ enum PowerType {
   POWER_TYPE_REACTIVE = 2;
 }
 
+// Enumerated power sources for an electrical component when issuing discharge
+// commands.
+//
+// This is needed only for discharge commands for hybrid inverters, which
+// typically have both PV and battery inputs added to them, and can use either
+// of those inputs as the power source.
+enum PowerSource {
+  // Default value.
+  //
+  // When used, the API service will automatically select
+  // `POWER_SOURCE_PV_PREFERRED` as the power source.
+  POWER_SOURCE_UNSPECIFIED = 0;
+
+  // When only the PV arrays should be used as the energy source.
+  // Commands will return an error if:
+  // - the requested power exceeds the limits of the PV arrays, or
+  // - the hybrid inverter does not have PV arrays added to it.
+  POWER_SOURCE_PV_ONLY = 1;
+
+  // When the PV arrays should be used as the preferred energy source,
+  // with any remaining power demand drawn from batteries.
+  // Commands will return an error if the requested power exceeds the limits of
+  // the PV arrays and batteries combined.
+  POWER_SOURCE_PV_PREFERRED = 2;
+
+  // When the batteries should be used as the preferred energy source,
+  // with any remaining power demand drawn from PV arrays where available.
+  // Commands will return an error if the requested power exceeds the limits of
+  // the batteries and PV arrays combined.
+  POWER_SOURCE_BATTERY_PREFERRED = 3;
+
+  // When only the batteries should be used as the energy source.
+  // Commands will return an error if:
+  // - the requested power exceeds the limits of the batteries, or
+  // - the hybrid inverter does not have batteries added to it.
+  POWER_SOURCE_BATTERY_ONLY = 4;
+}
+
 // Response codes for the `SetElectricalComponentPowerActive` and
 // `SetElectricalComponentPowerReactive` RPCs. These codes indicate the ongoing
 // status of the request to set the power of an electrical component.
@@ -561,6 +599,21 @@ message SetElectricalComponentPowerRequest {
 
   // The requested power type.
   PowerType power_type = 2;
+
+  // The power source for the electrical component.
+  //
+  // This field is used only if the following criteria are all met:
+  // - the requested power type is active power,
+  // - the requested power is negative, indicating discharge, and
+  // - the target electrical component is a hybrid inverter.
+  //
+  // If the above criteria are met, and this field is not specified and allowed
+  // to default, then the API service will use `POWER_SOURCE_PV_PREFERRED` as
+  // the power source.
+  //
+  // If the above criteria are not met, the API service will ignore this field
+  // and automatically select the power source.
+  PowerSource power_source = 5;
 
   // The requested power output.
   //
--- a/proto/frequenz/api/microgrid/v1alpha19/microgrid.proto
+++ b/proto/frequenz/api/microgrid/v1alpha19/microgrid.proto
@@ -497,6 +497,44 @@ enum PowerType {
   POWER_TYPE_REACTIVE = 2;
 }
 
+// Enumerated power sources for an electrical component when issuing discharge
+// commands.
+//
+// This is needed only for discharge commands for hybrid inverters, which
+// typically have both PV and battery inputs added to them, and can use either
+// of those inputs as the power source.
+enum PowerSource {
+  // Default value.
+  //
+  // When used, the API service will automatically select
+  // `POWER_SOURCE_PV_PREFERRED` as the power source.
+  POWER_SOURCE_UNSPECIFIED = 0;
+
+  // When only the PV arrays should be used as the energy source.
+  // Commands will return an error if:
+  // - the requested power exceeds the limits of the PV arrays, or
+  // - the hybrid inverter does not have PV arrays added to it.
+  POWER_SOURCE_PV_ONLY = 1;
+
+  // When the PV arrays should be used as the preferred energy source,
+  // with any remaining power demand drawn from batteries.
+  // Commands will return an error if the requested power exceeds the limits of
+  // the PV arrays and batteries combined.
+  POWER_SOURCE_PV_PREFERRED = 2;
+
+  // When the batteries should be used as the preferred energy source,
+  // with any remaining power demand drawn from PV arrays where available.
+  // Commands will return an error if the requested power exceeds the limits of
+  // the batteries and PV arrays combined.
+  POWER_SOURCE_BATTERY_PREFERRED = 3;
+
+  // When only the batteries should be used as the energy source.
+  // Commands will return an error if:
+  // - the requested power exceeds the limits of the batteries, or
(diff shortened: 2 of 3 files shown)


proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
```diff
@@ -497,6 +497,44 @@ enum PowerType {
   POWER_TYPE_REACTIVE = 2;
 }
 
+// The preferred power source for an electrical component when issuing a
```
florian-wagner-frequenz: Nit: this is not just the preferred power source, as it also allows for "XYZ_ONLY".

proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
```diff
@@ -562,6 +600,24 @@ message SetElectricalComponentPowerRequest {
   // The requested power type.
   PowerType power_type = 2;
 
+  // The preferred power source for the electrical component when setting a
```
florian-wagner-frequenz: Nit: this is not only preferred as it can also be directive.

proto/frequenz/api/microgrid/v1alpha19/microgrid.proto
```diff
@@ -579,13 +635,13 @@ message SetElectricalComponentPowerRequest {
   // - Positive values indicate inductive reactive power, i.e., the current
   //   is lagging the voltage when viewed from the perspective of the
   //   electrical component.
-  float power = 3;
+  float power = 4;
```
florian-wagner-frequenz, line 635: I am not sure whether we want to bump the field order. This is a breaking change in my eyes and should correspond to a release of a new alpha version. Otherwise it might be hard to distinguish this from places where nitrogend is running with the "old" v1alpha19 vs. the "new" v1alpha19 (you see what this is doing?)

florian-wagner-frequenz (commented):
LGTM in general, but l don't think we should be doing a breaking update to the Microgrid API without increasing to v1alpha20

llucax:
> LGTM in general, but l don't think we should be doing a breaking update to the Microgrid API without increasing to v1alpha20

v1alpha19 is unreleased

proto/frequenz/api/microgrid/v1alpha19/microgrid.proto
tiyash-basu-frequenz: Thanks for bringing it up, but there are a few of things to note here:

- We have not released anything beyond v0.18.0 yet, which means we have not yet released the `v1alpha19` package. We will do it in the future when we release v0.19.0. So this breaking change is valid now.
- If you look at our package hierarchy, you will notice that we have `v1alpha18` and `v1apha19` as two separate packages. No future release will break `v1alpha18`, and breaking changes should go only into `v1alpha19`. This is what this PR is also doing.
- Even after we make the next breaking release, users can still continue using the _older_ package `v1alpha18`, since we guarantee that it will never break.

So, this breaking change is not a problem. Breaking changes are valid to have _before_ making the release. :)

tiyash-basu-frequenz:
> but l don't think we should be doing a breaking update to the Microgrid API without increasing to v1alpha20

Also responded [here](https://github.com/frequenz-floss/frequenz-api-microgrid/pull/382#discussion_r2883594523).

proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
```diff
@@ -497,6 +497,44 @@ enum PowerType {
...
+  // Commands will return an error if the requested power exceeds the limits of
+  // the batteries and PV arrays combined.
+  POWER_SOURCE_BATTERY_PREFERRED = 3;
+
+  // When only the batteries should be used as the energy source.
+  // Commands will return an error if:
+  // - the requested power exceeds the limits of the batteries, or
+  // - the hybrid inverter does not have batteries added to it.
+  POWER_SOURCE_BATTERY_ONLY = 4;
+}
```
llucax, line 536: I find the ordering confusing, I would use:

POWER_SOURCE_PV_ONLY
POWER_SOURCE_PV_PREFERRED
POWER_SOURCE_BATTERY_ONLY
POWER_SOURCE_BATTERY_PREFERRED

or

POWER_SOURCE_PV_ONLY
POWER_SOURCE_BATTERY_ONLY
POWER_SOURCE_PV_PREFERRED
POWER_SOURCE_BATTERY_PREFERRED

But this "mirrored" order make my brain seasick 😆

llucax (dismissed):
Besides my small OC observation, LGTM.

proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
tiyash-basu-frequenz: Fixed.

proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
tiyash-basu-frequenz: Fixed.

proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
tiyash-basu-frequenz: The ordering I have now follows the PV->Battery direction, but I admit it is not clear enough. I will gladly change it if you want.

proto/frequenz/api/microgrid/v1alpha19/microgrid.proto
florian-wagner-frequenz: Ah, okay. I was not aware that v1alpha19 is not yet released, so then we'll be fine :)

florian-wagner-frequenz (approved):
LGTM, thank you for the context

proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
llucax: OK, I see, now that it has some logic, I guess my brain can cope better, so all good from my side to keep it as it is.

llucax (approved):

PR #414, opened 2026-04-29
Update frequenz-api-common submodule from v0.8.3 to v0.8.4

tiyash-basu-frequenz:


llucax (approved):

llucax (changes requested):
Wait, actually we need to also bump the python deps for `frequenz-api-common`.

llucax:
I suggest `"frequenz-api-common >= 0.8.4, < 1"`, as technically we should keep backwards compatibility, at least to the next major, in v1 we probably will want to remove all `v0alphaX` stuff.

tiyash-basu-frequenz:
Done, ready for another look.

llucax (approved):

PR #442, opened 2026-08-27
Add `bounds_token` and `ClearElectricalComponentBounds` RPC to v1alpha18/19

### LINKED ISSUE #441: Allow clients to override previously set bounds
### What's needed?

Clients often set bounds based upon the current telemetry data from the assets. Often, when the telemetry data indicates a change in the system, the previously set bounds need to be overwritten or withdrawn. The API does not allow that. The clients need to wait until the bounds expire when their lifetime ends. This is problematic if a client sets bounds for 15 minutes, and later has to no means to change it.

### Proposed solution

**Solution 1: Server-supplied token**

The [response body](https://github.com/frequenz-floss/frequenz-api-microgrid/blob/2acd3222514d3875c5ac5d3ca9acc67346f0173e/proto/frequenz/api/microgrid/v1alpha18/microgrid.proto#L478-L485) of the `AugmentElectricalComponentBounds` should contain a token that the client can use to withdraw or override previously issued bounds.

Once a client gets this token, they can re-use this token in the next `AugmentElectricalComponentBounds` request, as an optional parameter. This lets the Microgrid API service to overwrite the previous bound contributions (for which it had sent the token) with the new one. In response to the new request, the API service sends a token again (which should not be assumed to be the same as the previous token).

If a client wants to withdraw previously issued bounds completely, they make a new request with the token of those bounds and the `bounds` field set to an empty vector.

**Solution 2: Client-supplied token**

Solution 1 risks clients losing in-flight responses, thus losing the tokens they can use to overwrite older bound contributions.

An alternative here is to allow the client to send its own token to track its bounds. The `AugmentElectricalComponentBoundsRequest` message allows clients to optionally supply a token. A client may incorporate random bytes into this token, to avoid it clashing with other requests that the server receives.

This token functions similarly to solution 1 - it can be used to overwrite an older bound contribution by using it in a new request, and it can be used to clear older bound contributions.

### Use cases

- A client throttles a component on a telemetry reading, then needs to lift the throttle as soon as the reading recovers, rather than at the end of a lifetime it chose earlier.
- A client shutting down cleanly withdraws its bounds instead of leaving them to expire.
- Any client that wants a `request_lifetime` longer than a few seconds without giving up the ability to revise.

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
This PR adds a token-based mechanism for retiring electrical component bounds contributions early, applied identically to both the `v1alpha18` and `v1alpha19` proto packages.

Previously, bounds added via `AugmentElectricalComponentBounds` could only expire passively via `request_lifetime`.

The commit sequence here first introduces the `bounds_token` field on `AugmentElectricalComponentBoundsRequest`, letting a client tag its bounds contribution and later overwrite it by reusing the same token. It then adds a new `ClearElectricalComponentBounds` RPC that lets a client withdraw a tagged contribution outright, given the component, metric, and token.

## Non-Breaking Changes:
- **Add `bounds_token` field to `AugmentElectricalComponentBoundsRequest`**: Adds a new optional `bounds_token` field so clients can tag a bounds contribution and later overwrite it by reusing the token in a subsequent request, instead of waiting for `request_lifetime` to expire.
- **Add RPC to clear previously set bounds**: Introduces `ClearElectricalComponentBounds`, allowing a client to withdraw a bounds contribution (identified by component, metric, and `bounds_token`) before its lifetime elapses; missing/already-expired contributions are treated as a no-op.

## Dropped Alternative

We could have done this with with server-side-tokens too, but it has one drawback -  it risks clients losing in-flight responses. That is an inherently server-side token problem, and has an eventual consequence that defeats its purpose - if an in-flight response gets missed, then those bounds-contribution can never be overwritten or revoked. If the lifetime of such an outdated bounds-contribution is 15-minutes, then its effect lingers for long enough to affect microgrid performance.

The client-side-token approach resolves that problem. The only downside is that the client has to manage its own token. It also offers a potentially more powerful interface to manage its bounds, since clients can now track their bounds-contributions themselves.

closes #441

### DIFF
--- a/proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
+++ b/proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
@@ -136,6 +136,12 @@ service Microgrid {
   // which the bounds will stay in effect. If no duration is provided, then the
   // bounds will be removed after a default duration of 5 seconds.
   //
+  // The RPC offers two ways to retire bounds, both via its request body:
+  // - by providing a `request_lifetime` duration, after which the bounds will
+  //   be automatically removed, or
+  // - by providing a `bounds_token`, which can be re-used in a subsequent
+  //   request to immediately overwrite the bounds added by an older request.
+  //
   // Inclusion bounds give the range that the system will try to keep the
   // metric within. If the metric goes outside of these bounds, the system will
   // try to bring it back within the bounds.
@@ -472,7 +478,30 @@ message AugmentElectricalComponentBoundsRequest {
   // This duration has to be between 5 seconds and 15 minutes (including both
   // limits), otherwise the request will be rejected.
   // If not provided, it defaults to 5s.
+  //
+  // Note that the bounds may also expire before this duration, if the client
+  // provides the same `bounds_token` in a subsequent request for the same
+  // electrical component and metric.
   optional uint64 request_lifetime = 4;
+
+  // A client-generated token, used to track the bounds added by this request.
+  //
+  // It can be re-used in a subsequent request for the same electrical component
+  // and metric to immediately overwrite the bounds added by this request, and
+  // remove them from the overall bounds for the given metric.
+  //
+  // The service should allow the same token to be used for different electrical
+  // component IDs and metrics. This means that clients can expect that using
+  // the same token in a subsequent request for a different electrical component
+  // or metric will _not_ overwrite the bounds added by this request.
+  //
+  // If there is no bound contribution currently active for the given token,
+  // electrical component ID, and metric, then the request will be treated as a
+  // new one.
+  //
+  // Clients are strongly recommended to incorporate randomness in the token to
+  // avoid collisions, e.g. by using a random 64-bit value.
+  optional string bounds_token = 5;
 }
 
 // Response message for the RPC `AugmentElectricalComponentBounds`.
@@ -481,6 +510,11 @@ message AugmentElectricalComponentBoundsResponse {
   // After this timestamp, the electrical component bounds added by this request
   // will be removed. By default, this timestamp will be set to the current
   // time plus 60 seconds.
+  //
+  // Note that if the client provides the same `bounds_token` in a subsequent
+  // request for the same electrical component and metric, then the bounds added
+  // by this request will be removed immediately, and the `valid_until_time`
+  // timestamp will not be relevant anymore.
   google.protobuf.Timestamp valid_until_time = 1;
 }
 
--- a/proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
+++ b/proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
@@ -171,6 +171,18 @@ service Microgrid {
   rpc AugmentElectricalComponentBounds(AugmentElectricalComponentBoundsRequest)
     returns (AugmentElectricalComponentBoundsResponse);
 
+  // This RPC allows clients to remove bounds that were previously added using
+  // the `AugmentElectricalComponentBounds` RPC. Clients can specify the
+  // electrical component ID, metric, and token to identify the bounds to be
+  // cleared.
+  //
+  // If the specified bounds exist, they will be removed from the overall bounds
+  // for the given metric of the electrical component.
+  // If the specified bounds do not exist, the request will be treated as a
+  // no-op.
+  rpc ClearElectricalComponentBounds(ClearElectricalComponentBoundsRequest)
+    returns (google.protobuf.Empty);
+
   // Sets the active or reactive power output of a electrical component to a
   // specified target.
   // This RPC allows setting either active or reactive power in a single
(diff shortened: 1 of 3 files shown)


proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
```diff
@@ -472,7 +490,30 @@ message AugmentElectricalComponentBoundsRequest {
   // This duration has to be between 5 seconds and 15 minutes (including both
   // limits), otherwise the request will be rejected.
   // If not provided, it defaults to 5s.
+  //
+  // Note that the bounds may also expire before this duration, if the client
+  // provides the same `bounds_token` in a subsequent request for the same
+  // electrical component and metric.
   optional uint64 request_lifetime = 4;
+
+  // A client-generated token, used to track the bounds added by this request.
```
florian-wagner-frequenz: Do we do anything to prevent collision between clients?
I assume no and this is mainly "protect against stupidity, not malice", right?

proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
```diff
@@ -472,7 +490,30 @@ message AugmentElectricalComponentBoundsRequest {
...
+  //
+  // It can be re-used in a subsequent request for the same electrical component
+  // and metric to immediately overwrite the bounds added by this request, and
+  // remove them from the overall bounds for the given metric.
+  //
+  // The service should allow the same token to be used for different electrical
+  // component IDs and metrics. This means that clients can expect that using
+  // the same token in a subsequent request for a different electrical component
+  // or metric will _not_ overwrite the bounds added by this request.
```
florian-wagner-frequenz: I assume what you're saying here:
"Bounds are tracked per component ID or metric, so you can re-use the token without wiping the place where you first used it. At the same time if you issue a command to an already bounded component ID and metric, you can update, i.e. widen that bound."
Is that understanding correct?

proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
```diff
@@ -481,9 +522,34 @@ message AugmentElectricalComponentBoundsResponse {
   // After this timestamp, the electrical component bounds added by this request
   // will be removed. By default, this timestamp will be set to the current
   // time plus 60 seconds.
+  //
+  // Note that if the client provides the same `bounds_token` in a subsequent
+  // request for the same electrical component and metric, then the bounds added
+  // by this request will be removed immediately, and the `valid_until_time`
+  // timestamp will not be relevant anymore.
```
florian-wagner-frequenz: Instead the new valid_until_time timestamp will be in effect for the overwritten bounds?

proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
llucax: Oh, interesting, *client generated*. This means different clients can mess (intentionally if they know the token, or unintentionally if they don't but there is a *collision*) with each other set bounds, right?

Is this really what we want? I thought the API would have been that the server creates a hash-like token and gives it back to the client to optionally be used to override/clear a set bounds.

Two very different use cases I guess, and I guess this being *client generated* is intentional and what we need, but better to be sure :)

llucax (commented):
Other than the question here, LGTM.

llucax (approved):
Approving in case it is intentional and you want to merge.

hannah-stevenson-frequenz (approved):

proto/frequenz/api/microgrid/v1alpha18/microgrid.proto
ahmet-erdem-frequenz: I also thought that the token would be generated on the server-side and passed on to the client. Allowing the clients to refer to their bounds to be overwritten.

tiyash-basu-frequenz: I eventually decided to go for a client-side tokens.

The server-side-token approach is apparently simpler for the clients, but it risks clients losing in-flight responses. That is an inherently server-side token problem, and has an eventual consequence that defeats its purpose - if an in-flight response gets missed, then those bounds-contribution can never be overwritten or revoked. If the lifetime of such an outdated bounds-contribution is 15-minutes, then its effect lingers for long enough to affect microgrid performance.

The client-side token approach resolves that problem. The only downside is that the client has to manage its own token. It also offers a potentially more powerful interface to manage its bounds, since clients can now track their bounds-contriobutions themselves.

I had mentioned this in the issue description, but I guess that wasn't enough. I will update the PR description now.

I will wait to hear your thoughts on the decision, and keep the PR unmerged until then.

ahmet-erdem-frequenz (approved):
LGTM! 

I think I understand the risk of client-side token now.

llucax:
https://github.com/frequenz-floss/frequenz-api-microgrid/pull/442#discussion_r3892763114

> The server-side-token approach is apparently simpler for the clients, but it risks clients losing in-flight responses. That is an inherently server-side token problem, and has an eventual consequence that defeats its purpose - if an in-flight response gets missed, then those bounds-contribution can never be overwritten or revoked.

I don't understand what do you mean by this? How can a response get missed? A disconnection? All CRUD systems work with the server returning an ID, I'm surprised this risk is big enough to go with the client-side token, which also have non-trivial implications, like hijacking someone else's request or collisions.

I think we should think well about the full implications of both approaches before doing a release.

llucax:
Just for the records, we talked about the trade-offs and since the docs already explain the risks of collisions, we should be good, and the client side token provides some extra flexibility and also should be a NOP when unused (against always returning a token, even if nobody asked for it).

